PR #10, opened 2025-02-03
chore: swap blockenv values instead of cloning 

### LINKED ISSUE #9: [Feature] Remove blockenv clone on system call

https://github.com/alloy-rs/evm/blob/0384bf0ad513909fc51422f69734343fe513ddb6/crates/evm/src/eth.rs#L65-L71

clones the blockenv in order to update to fields,

instead we can simply swap those fields before and after transact

### Additional context

_No response_

Evalir:
## Motivation

Closes #9

## Solution

swaps the individual blockenv values instead of cloning the entire block env.

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/evm/src/eth.rs
+++ b/crates/evm/src/eth.rs
@@ -66,16 +66,16 @@ impl<EXT, DB: Database> Evm for EthEvm<'_, EXT, DB> {
         let mut basefee = U256::ZERO;
 
         // ensure the block gas limit is >= the tx
-        std::mem::swap(&mut self.block_mut().gas_limit, &mut gas_limit);
+        core::mem::swap(&mut self.block_mut().gas_limit, &mut gas_limit);
         // disable the base fee check for this call by setting the base fee to zero
-        std::mem::swap(&mut self.block_mut().basefee, &mut basefee);
+        core::mem::swap(&mut self.block_mut().basefee, &mut basefee);
 
         let res = self.0.transact();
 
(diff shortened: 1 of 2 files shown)


mattsse (approved):
lgtm,

this should be fine because immutable during execution, right @klkvr

crates/evm/src/eth.rs
```diff
@@ -62,18 +62,20 @@ impl<EXT, DB: Database> Evm for EthEvm<'_, EXT, DB> {
...
+        let mut gas_limit = U256::from(self.tx().gas_limit);
+        let mut basefee = U256::ZERO;
 
         // ensure the block gas limit is >= the tx
-        self.block_mut().gas_limit = U256::from(self.tx().gas_limit);
-
+        std::mem::swap(&mut self.block_mut().gas_limit, &mut gas_limit);
         // disable the base fee check for this call by setting the base fee to zero
-        self.block_mut().basefee = U256::ZERO;
+        std::mem::swap(&mut self.block_mut().basefee, &mut basefee);
```
mattsse, line 71: these should be `core::mem`

klkvr (approved):
thank you! can we do the same for op as well?

Evalir:
sg @klkvr , done in https://github.com/alloy-rs/evm/pull/10/commits/1e987b58fe76d63d1f7fef12899f5c4d3f10de3f

PR #11, opened 2025-02-07
feat: add error traits and ATs

klkvr:
## Motivation

Extracts some API changes from https://github.com/alloy-rs/evm/pull/7 to make it easier to prototype with reth in scope of https://github.com/paradigmxyz/reth/pull/14021

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/evm/src/evm.rs
```diff
@@ -14,12 +21,17 @@ use revm::{
 /// Executing a transaction will return the outcome of the transaction.
 pub trait Evm {
     /// Database type held by the EVM.
-    type DB: Database;
+    type DB;
```
mattsse, line 24: oh that's kinda nice

mattsse (approved):

crates/evm/src/evm.rs
klkvr: it's actually just a workaround for weird trait bounds issue :/ we still require `DB: Database` on EvmFactory

klkvr:
ci is failing because current revm feature-gates `Error` impls behind `std` so holding this off

mattsse:
since this is a new release basically revm ideally bumps msrv so we can use core

emhane:
> ci is failing because current revm feature-gates `Error` impls behind `std` so holding this off

is it possible to ship a std only version of this first of all, so I can base re-impl of https://github.com/paradigmxyz/reth/pull/14124 on main with https://github.com/paradigmxyz/reth/pull/14021 merged? that way we can work in parallel and be more efficient. alt is there a revm issue for making `Error` impls `no_std` compatible?

PR #12, opened 2025-02-11
chore: relax a generic trait

rakita:
Precompile generic used for `EthEvm`, by default it will be `EthPrecompiles`


crates/evm/src/eth.rs
```diff
@@ -8,21 +8,37 @@ use core::{
...
 
 /// The Ethereum EVM context type.
 pub type EthEvmContext<DB> = Context<BlockEnv, TxEnv, CfgEnv, DB>;
 
 /// Ethereum EVM implementation.
 #[expect(missing_debug_implementations)]
-pub struct EthEvm<DB: Database, I>(MainnetEvm<EthEvmContext<DB>, I>);
+pub struct EthEvm<DB: Database, I, PRECOMPILE = EthPrecompiles<EthEvmContext<DB>>>(
+    RevmEvm<EthEvmContext<DB>, I, EthInstructions<EthInterpreter, EthEvmContext<DB>>, PRECOMPILE>,
+);
```
mattsse, line 27: is there no mainnet evm anymore?

isn't this intended as a wrapper around the plain L1 evm?

imo this shouldn't require this much setup

mattsse: ah I think this is so that this specific `EthEvm` type can be reused

mattsse: this will likely be a common use case so worth thinking how we can simplify this setup out of the box

rakita: > ah I think this is so that this specific EthEvm type can be reused

Yes, so we can change precompiles as in reth precompile example.

As the default value for Precompiles is EthPrecompile it will be the same as the previous Mainnet type.

We could use `Box<dyn Precompiles>` that would remove generic from it.

PR #17, opened 2025-02-19
feat: add `EvmError::try_into_invalid_tx_err`

klkvr:
## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/evm/src/error.rs
+++ b/crates/evm/src/error.rs
@@ -23,13 +23,16 @@ impl InvalidTxError for InvalidTransaction {
 ///
 /// If caller occurs a error different from [`EvmError::InvalidTransaction`], it should most likely
 /// be treated as fatal error flagging some EVM misconfiguration.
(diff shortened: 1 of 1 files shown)


crates/evm/src/error.rs
```diff
@@ -23,14 +23,17 @@ impl InvalidTxError for InvalidTransaction {
 ///
 /// If caller occurs a error different from [`EvmError::InvalidTransaction`], it should most likely
 /// be treated as fatal error flagging some EVM misconfiguration.
-pub trait EvmError: Error + Send + Sync + 'static {
+pub trait EvmError: Sized + Error + Send + Sync + 'static {
```
mattsse, line 26: imo we can expect this

mattsse (approved):
makes sense

can't think of a usecase where this wouldn't be sized

PR #19, opened 2025-02-21
feat: `TxEnv` conversion traits

klkvr:
## Motivation

ref https://github.com/paradigmxyz/reth/issues/14551

This PR introduces two traits:
- `IntoTxEnv` - core conversion trait performing simple conversion from a `Tx` to `TxEnv` type. `impl IntoTxEnv<Self::Tx>` is now used as input on `Evm::transact` allowing to extend a set of transaction environment types that can be passed 
- `FromRecoveredTx` - a helper trait to provide `IntoTxEnv` implementations for `Recovered<Tx>`

This should allow us to cover all of the usecase as all of `TxEnv`, `Recovered<Tx>`, `Recovered<&Tx>` are now valid inputs for Evm, and users are still free to define `IntoTxEnv` for any external types.

One required hack is that any Evm impl has to define for `IntoTxEnv<Self>` for the `Evm::Tx` which is not perfect but is not that bad generally as we already require similar `AsRef<Self>` implementations to be provided in reth

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/evm/src/evm.rs
+++ b/crates/evm/src/evm.rs
@@ -27,6 +27,10 @@ pub trait Evm {
     /// Database type held by the EVM.
     type DB;
     /// The transaction object that the EVM will execute.
+    ///
+    /// Implementations are expected to rely on a single entrypoint for transaction execution such
+    /// as [`revm::context::TxEnv`]. The actual set of valid inputs is not limited by allowing to
+    /// provide any [`IntoTxEnv`] implementation for [`Evm::transact`] method.
     type Tx;
     /// Error type returned by EVM. Contains either errors related to invalid transactions or
     /// internal irrecoverable execution errors.
--- a/crates/evm/src/evm.rs
+++ b/crates/evm/src/evm.rs
@@ -1,6 +1,6 @@
 //! Abstraction over EVM.
 
-use crate::EvmError;
+use crate::{EvmError, IntoTxEnv};
 use alloy_primitives::{Address, Bytes};
 use core::error::Error;
 use revm::{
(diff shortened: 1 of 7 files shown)


crates/evm/src/evm.rs
```diff
@@ -40,7 +40,19 @@ pub trait Evm {
     fn block(&self) -> &BlockEnv;
 
     /// Executes a transaction and returns the outcome.
-    fn transact(&mut self, tx: Self::Tx) -> Result<ResultAndState<Self::HaltReason>, Self::Error>;
+    fn transact_raw(
```
mattsse, line 43: can we provide some additional context in docs here and on the AT tx that this is supposed to the transaction type (previously called TxEnv) the evm operates on

crates/evm/src/tx.rs
```diff
@@ -0,0 +1,80 @@
...
+    fn into_tx_env(self) -> Self {
+        self
+    }
+}
+
+/// Helper user-facing trait to allow implementing [`IntoTxEnv`] on instances of [`Recovered`].
+pub trait FromRecoveredTx<Tx> {
+    /// Builds a `TxEnv` from a transaction and a sender address.
+    fn from_recovered_tx(tx: &Tx, sender: Address) -> Self;
+}
```
mattsse, line 30: ah nice

crates/evm/src/tx.rs
```diff
@@ -0,0 +1,80 @@
...
+        self
+    }
+}
+
+#[cfg(feature = "op")]
+impl<T: revm::context::Transaction> IntoTxEnv<Self> for revm_optimism::OpTransaction<T> {
+    fn into_tx_env(self) -> Self {
+        self
+    }
+}
```
mattsse, line 24: can we experiment with this for the OP tx conversion:

https://github.com/paradigmxyz/reth/blob/86e8a2a245490a1503c3c05c246f63f3d48e0175/crates/optimism/payload/src/payload.rs#L32-L34

for example when we already have the encoded raw tx data we could use this to optimize this step.

I believe this is just a matter of impl FromRecoveredTx for `WithEncoded<OpTx>` 

or  `IntoTxEnv` for `WithEncoded<Recovered<OpTx>>`

crates/evm/src/evm.rs
```diff
@@ -40,7 +40,19 @@ pub trait Evm {
...
+        &mut self,
+        tx: Self::Tx,
+    ) -> Result<ResultAndState<Self::HaltReason>, Self::Error>;
+
+    /// Same as [`Evm::transact_raw`], but takes a [`IntoTxEnv`] implementation, thus allowing to
+    /// support transacting with an external type.
+    fn transact(
+        &mut self,
+        tx: impl IntoTxEnv<Self::Tx>,
```
mattsse, line 52: this is also impl for `Self::Tx`, so is there a technically reason, besides avoiding the into call why we need this extra function?

mattsse (changes requested):
doc nits and some qs

crates/evm/src/evm.rs
klkvr: yeah this is just to avoid an additional into call

mattsse: reasonable

mattsse (approved):

PR #22, opened 2025-03-03
refactor: embed EvmEnv into abstraction

klkvr:
## Motivation

It is useful to have a way to convert `Evm` back into `EvmEnv` once we're done with it. This requires removing `Input` generic on `EvmFactory` in favor of `Spec` AT. This means we can't define multiple inputs for the same factory type which is imo fine as Evm still takes fixed type as input, and custom inputs can be solved with additional traits (e.g like we've done it with `ConfigureEvmEnv` in reth)

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mattsse (approved):
makes sense

PR #26, opened 2025-03-06
feat: add Ethereum block execution

klkvr:
- Moves Ethereum `BlockExecutor` implementation from reth along with eth-specific state transitions (`dao_fork`, `eip6110`).
- To allow plugging custom receipt and transaction type, generic `ReceiptBuilder` is introduced. Basic implementation of `AlloyReceiptBuilder` is provided and will be used as default generic for the builder.
- `EthExecutorSpec` trait is introduced, extending `EthereumHardforks` with additional method to get `deposit_contract_address` for eip6110 implementation. I've added basic `EthSpec` struct for the sake of providing an implementation for external users.
- I've changed naming to use `spec`/Spec` vs `chain_spec`/`ChainSpec` to avoid possible naming conflicts with reth in future


mattsse (approved):
good stuff

PR #27, opened 2025-03-06
feat: add state hook closure fns

mattsse:
adds helper fns that make it possible to invoke the state hook with a closure so that we're not required to always collect the evmstate and can do that in a closure instead

can be used after #26


klkvr (approved):
right, this makes sense

PR #28, opened 2025-03-07
feat: `BlockExecutorFactory`

klkvr:
## Motivation

Moves `BlockExecutionStrategyFactory` from reth with changed name and a bit different semantics.

I've decided to remove GAT for `Strategy`/`Executor` in favor of using RPIT with a helper trait. It solves the redundancy of writing long `impl BlockExecutor<Evm = Self::Evm<'a, DB, I>, Transaction = Self::Tranaction, ...>` and allows to make `BlockExecutorFactory` impls smaller as GAT definition is pretty big.

We could do the same for Evm as this GAT is not ideal as well, https://github.com/alloy-rs/evm/blob/0340109184517986eb8bd7f5ac20d67513f80920/crates/evm/src/evm.rs#L116-L122

It is also impractical to bound because non-lifetime binders (`Evm: for<DB, I>: MyTrait`) are not supported. However, it is still useful in scope of block executors which might depend on a concrete type of `Evm` produced by `EvmFactory` for e.g using some custom helpers defined on it 

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/evm/src/block/mod.rs
+++ b/crates/evm/src/block/mod.rs
@@ -1,10 +1,12 @@
 //! Block execution abstraction.
 
-use crate::Evm;
+use crate::{Evm, EvmFactory, Database};
 use alloc::{boxed::Box, vec::Vec};
 use alloy_consensus::transaction::Recovered;
 use alloy_eips::eip7685::Requests;
-use revm::context::result::ExecutionResult;
+use revm::{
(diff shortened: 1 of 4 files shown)


mattsse (approved):
lgtm!

PR #41, opened 2025-03-14
feat: add toggle for inspector

### LINKED ISSUE #38: [Feature] Support improved block level inspecting

currently the `Evm` instance doesn't expose a way to enable/disable the inspector once the instance has been created.

this has some drawbacks for block level tracing which can consists of:

* execute txs in a block until target tx then inspect target tx
* inspect all txs ina block

the former currently requires creating a new evm instance.

we can fix this with a simple toggle since we have a bool for this, so that the former use case works like:

1. create evm with inspector
2. disable inspecting
3. execute all txs until the tx
4. enable inspector
5. transact with inspector

this isn't super intuitive but should work and we can likely make this a lot easier via some extension type/trait

# TODO
* add enable/disable_inspector functions

### Additional context

_No response_

varun-doshi:
## Motivation

Closes #38 
currently the Evm instance doesn't expose a way to enable/disable the inspector once the instance has been created.

this has some drawbacks for block level tracing which can consists of:

execute txs in a block until target tx then inspect target tx
inspect all txs ina block

## Solution

added enable/disable_inspector functions

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


crates/evm/src/eth/mod.rs
```diff
@@ -83,6 +83,16 @@ impl<DB: Database, I, PRECOMPILE> EthEvm<DB, I, PRECOMPILE> {
...
+
+    /// Disables the inspector.
+    pub fn enable_inspector(&mut self) {
+        self.inspect = true;
+    }
+
+    /// Disables the inspector.
+    pub fn disable_inspector(&mut self) {
+        self.inspect = false;
+    }
```
mattsse, line 95: we also want another fn that accepts a bool

and we need to move this to the evm trait, then these two books can call the one that accepts the bool by default so that the impls only have to implement one trait function

mattsse (changes requested):

Issue #40, opened 2025-03-13
[Feature] Add helper trait for deriving optxenv from `WithEncoded`


mattsse:
### Describe the feature you would like

the optimism txenv requires that we get the 2718 encoded bytes,

we often have this already and could provide this as input to create the txenv,
this like needs an additional trait or impl for `WithEncoded`

@klkvr

klkvr:
For this we'd need 2 changes:
1. Currently, `BlockExecutor::execute_transaction` takes only `Recovered<Self::Transaction>`. We'd need to relax this to take somthing that can be both converted into `Evm::TxEnv` and to a transaction e.g smth like `AsRef<Recovered<&Tx>> + IntoTxEnv<Self::Tx>` or a separate trait. Need to make sure that this doesn't introduce any unnecesarry cloning during conversions
2. Add a helper trait for `WithEncoded<T>` conversions like we've done for `Recovered`: https://github.com/alloy-rs/evm/blob/685caa38cd2532363b82fdd1733720ab1199c33a/crates/evm/src/tx.rs#L41-L45
    i.e user should be able to define `FromTxWithEncoded<MyTx> for OpTransaction<TxEnv>` thus making `WithEncoded<MyTx>` to implement `IntoTxEnv<OpTransaction<TxEnv>>`

lean-apple:
Hello, may I work on it ?

PR #42, opened 2025-03-16
feat: add helper trait for deriving `TxEnv` from `WithEncoded`

lean-apple:
## Motivation

Closes https://github.com/alloy-rs/evm/issues/40.

## Solution

- Added `FromEncodedTx<Tx>` helper trait to build transaction environments from encoded bytes
- Implemented `IntoTxEnv<TxEnv>` for `WithEncoded<T>` types
- Added `AsRecoveredTx<T>` trait to access recovered transactions
- Relaxed `BlockExecutor::execute_transaction` and `execute_transaction_with_result_closure` to accept any type implementing both `IntoTxEnv<TxEnv>` and `AsRecoveredTx<Self::Transaction>`

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Breaking changes


crates/evm/src/tx.rs
```diff
@@ -50,6 +51,46 @@ impl<T, TxEnv: FromRecoveredTx<T>> IntoTxEnv<TxEnv> for &Recovered<T> {
...
+        Recovered::new_unchecked(self.inner(), self.signer())
+    }
+}
+
+/// Helper user-facing trait to allow implementing [`IntoTxEnv`] on instances of [`WithEncoded`].
+/// This allows creating transaction environments directly from EIP-2718 encoded bytes.
+pub trait FromEncodedTx<Tx> {
+    /// Builds a `TxEnv` from encoded transaction bytes.
+    fn from_encoded_tx(encoded: &[u8]) -> Self;
+}
```
klkvr, line 71: we also need this to have access to transaction itself and its sender, so I believe signature would need to be something like `from_encoded_tx(tx, sender, encoded)`

crates/evm/src/tx.rs
```diff
@@ -50,6 +51,46 @@ impl<T, TxEnv: FromRecoveredTx<T>> IntoTxEnv<TxEnv> for &Recovered<T> {
     }
 }
 
+/// Helper user-facing trait to allow retrieving a recovered transaction reference.
+pub trait AsRecoveredTx<T> {
+    /// Returns a reference to the recovered transaction.
+    fn as_recovered(&self) -> Recovered<&T>;
+}
```
klkvr, line 58: We should implement this for `WithEncoded<Recovered<T>>` out of the box so that it can be passed into block executors

crates/evm/src/eth/block.rs
```diff
@@ -104,12 +106,14 @@ where
 
     fn execute_transaction_with_result_closure(
         &mut self,
-        tx: Recovered<&R::Transaction>,
+        tx: impl IntoTxEnv<TxEnv> + AsRecoveredTx<Self::Transaction>,
```
klkvr, line 109: we should now use `IntoTxEnv` when converting to `tx_env`

the assumption here is that `IntoTxEnv` will either be the same as `.as_recovered_tx().into_tx_env()` or more optimized (like in case with `WithEncoded`)

klkvr (commented):
this is already looking good, left some comments

crates/evm/src/eth/block.rs
lean-apple: I tried to use directly `().into_tx_env()` on `tx.as_recovered_tx()` after the gas checks, 

yet I'm not sure it's possible to use it directly without having a new impl like this 
```
impl<R> FromRecoveredTx<R::Transaction> for TxEnv 
where 
    R: eth::receipt_builder::ReceiptBuilder,
    R::Transaction: Transaction,
{
    fn from_recovered_tx(tx: &R::Transaction, sender: Address) -> Self {
   }
}
```
Should I add to enable the conversion ?

klkvr: we don't want to use the `into_tx_env` on `tx._recovered_tx()` as in case of `WithEncoded` we'll lose optimization that `WithEncoded` gives us

instead it should be something like
```
let tx = tx.as_recovered();
let tx_env = tx.into_tx_env();
```

or maybe we should even add a separate trait to avoid having to drop value before `into_tx_env` call

mattsse (approved):
lgtm!

PR #44, opened 2025-03-19
fix: correctly handle state after system calls

klkvr:
## Motivation

There's an edge case when system contract is set to block beneficiary, so we can't simply remove it from the state. The assumption here is that there will not be any state changes to beneficiary due to fee accounting, and we can safely keep it in the state.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/evm/src/eth/mod.rs
+++ b/crates/evm/src/eth/mod.rs
@@ -167,7 +167,7 @@ where
         // disable the nonce check
         core::mem::swap(&mut self.cfg.disable_nonce_check, &mut disable_nonce_check);
 
-        let res = self.transact(tx);
+        let mut res = self.transact(tx);
 
         // swap back to the previous gas limit
         core::mem::swap(&mut self.block.gas_limit, &mut gas_limit);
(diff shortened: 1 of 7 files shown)


rakita (approved):
lgtm

crates/evm/src/eth/mod.rs
```diff
@@ -176,6 +176,11 @@ where
         // swap back to the previous nonce check flag
         core::mem::swap(&mut self.cfg.disable_nonce_check, &mut disable_nonce_check);
 
+        // NOTE: We assume that only the contract storage is modified.
+        if let Ok(res) = &mut res {
+            res.state.retain(|addr, _| *addr == contract);
```
mattsse, line 181: can we add this to the trait docs as well because this isn't obvious.

I also want some additional context here, why we assume this and what happens if we don't do this

mattsse (changes requested):

mattsse (approved):

PR #45, opened 2025-03-21
fix(isthmus): Assign deposit tx type to system call txs

clabby:
## Overview

Assigns the deposit tx type to system call transactions, so that `op-revm` skips transaction fee validation for them. The operator fee now can charge a non-zero fee even if the base fee is zero, and this bypasses the check and the attempted charge for pre-block syscalls.

http://github.com/bluealloy/revm/blob/080c29a2aef640063c7bcb30834ef7e998b8efe2/crates/optimism/src/handler.rs#L89-L90

### DIFF
--- a/crates/op-evm/src/lib.rs
+++ b/crates/op-evm/src/lib.rs
@@ -16,6 +16,7 @@ use core::{
     fmt::Debug,
     ops::{Deref, DerefMut},
 };
+use op_alloy_consensus::OpTxType;
 use op_revm::{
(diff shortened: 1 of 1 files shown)


crates/op-evm/src/lib.rs
```diff
@@ -145,7 +146,7 @@ where
                 // blob fields can be None for this tx
                 blob_hashes: Vec::new(),
                 max_fee_per_blob_gas: 0,
-                tx_type: 0,
+                tx_type: OpTxType::Deposit as u8,
```
mattsse, line 149: I see, system tx should have system tx type

mattsse (approved):

PR #50, opened 2025-03-31
added execute_block

### LINKED ISSUE #49: [Feature] Add helper function to execute an entire block via BlockExecutor

```
        strategy.apply_pre_execution_changes()?;
        for tx in block.transactions_recovered() {
            strategy.execute_transaction(tx)?;
        }
        let result = strategy.apply_post_execution_changes()?;
```

https://github.com/paradigmxyz/reth/blob/409e1db7f2aae87fcdd40f41ce6fd7d16518c807/crates/evm/src/execute.rs#L436-L440

this can be a new function with a default impl that accepts an impl IntoIterator and does all of this

### Additional context

_No response_

Rimeeeeee:
fixes #49

### DIFF
--- a/crates/evm/src/block/mod.rs
+++ b/crates/evm/src/block/mod.rs
@@ -112,22 +112,22 @@ pub trait BlockExecutor {
 
     /// Exposes mutable reference to EVM.
     fn evm_mut(&mut self) -> &mut Self::Evm;
-    /// Executes a block of transactions, applying pre and post execution changes.
+
+    /// Executes all transactions in a block, applying pre and post execution changes.
     fn execute_block(
-        self,
+        mut self,
         transactions: impl IntoIterator<Item = impl ExecutableTx<Self>>,
     ) -> Result<BlockExecutionResult<Self::Receipt>, BlockExecutionError>
     where
         Self: Sized,
     {
-        let mut this = self;
-        this.apply_pre_execution_changes()?;
+        self.apply_pre_execution_changes()?;
 
         for tx in transactions {
-            this.execute_transaction(tx)?;
+            self.execute_transaction(tx)?;
         }
 
-        this.apply_post_execution_changes()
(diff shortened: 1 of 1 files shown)


Rimeeeeee:
@mattsse , is this ok?

crates/evm/src/block/mod.rs
```diff
@@ -112,6 +112,23 @@ pub trait BlockExecutor {
...
+    /// Executes a block of transactions, applying pre and post execution changes.
+    fn execute_block(
+        self,
+        transactions: impl IntoIterator<Item = impl ExecutableTx<Self>>,
+    ) -> Result<BlockExecutionResult<Self::Receipt>, BlockExecutionError>
+    where
+        Self: Sized,
+    {
+        let mut this = self;
+        this.apply_pre_execution_changes()?;
```
mattsse, line 124: we likely dont need the `this` var and can just operate on self?

crates/evm/src/block/mod.rs
```diff
@@ -112,6 +112,23 @@ pub trait BlockExecutor {
 
     /// Exposes mutable reference to EVM.
     fn evm_mut(&mut self) -> &mut Self::Evm;
+    /// Executes a block of transactions, applying pre and post execution changes.
```
mattsse, line 115: needs new line here

mattsse, line 115:
```suggestion
    /// Executes all transactions in a block, applying pre and post execution changes.
```

mattsse (changes requested):
pedantic nits

crates/evm/src/block/mod.rs
Rimeeeeee: sureeee

Rimeeeeee:
> pedantic nits

@mattsse , done :)

mattsse (approved):
lgtm, ty

Issue #52, opened 2025-04-01
[Bug] inspector in create_evm()


PowVT:
### What version of Alloy are you on?

alloy-evm = { version = "0.2.0", default-features = false }

### Describe the bug

Curious why is there a `inspect: bool`?

```
pub trait Evm {
/// Determines whether additional transactions should be inspected or not.
///
/// See also [`EvmFactory::create_evm_with_inspector`].
fn set_inspector_enabled(&mut self, enabled: bool);
```

When using the evm factory I think there could be just one function `create_evm()` where can can pass in any inspector not just the NoOpInspector. Then toggle it on/off with the `enable_inspector()` methods. Currently I dont think `enable_inspector()` is doing anything since NoOpInspsector is in the function definition.

PowVT:
> it does, it then inspects but using the Noop, which isn't really what you want.
> 
> but this allows us the get rid of the inspector trait that is enshrined in the revm evm type

Thanks, does this mean the `enable_inspector()` toggle does nothing currently?

PowVT:
ok thanks, yeah I am curious if we can remove the Noop inspector to be able to pass in a custom inspector there?

PR #53, opened 2025-04-01
rm create_evm_with_inspector fn

PowVT:
Change regarding issue #52 

Seems redundant to have inspector flag passed to `EthEvm::new()` and also have another function which creates the same evm with an inspector. In the current design the create_evm() inspector can not be enabled via the enable_inspector toggle since it is a `NoOpInspector`.

## Motivation

Say I want to call `self.evm.enable_inspector()` in `BlockExecutor:: execute_transaction_with_result_closure()` and then transact the tx... In current design the inspector is `NoOpInspector` so a custom inspector could not be used in `execute_transaction_with_result_closure()`.

### DIFF
--- a/crates/op-evm/src/block/mod.rs
+++ b/crates/op-evm/src/block/mod.rs
@@ -314,7 +314,7 @@ mod tests {
     use alloy_evm::EvmEnv;
     use alloy_primitives::{Address, PrimitiveSignature as Signature};
     use op_alloy_consensus::OpTxEnvelope;
-    use revm::database::{CacheDB, EmptyDB};
+    use revm::{database::{CacheDB, EmptyDB}, inspector::NoOpInspector};
 
     use super::*;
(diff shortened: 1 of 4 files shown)


crates/op-evm/src/block/mod.rs
```diff
@@ -326,7 +326,7 @@ mod tests {
             OpEvmFactory::default(),
         );
         let mut db = State::builder().with_database(CacheDB::<EmptyDB>::default()).build();
-        let evm = executor_factory.evm_factory.create_evm(&mut db, EvmEnv::default());
+        let evm = executor_factory.evm_factory.create_evm(&mut db, EvmEnv::default(), NoOpInspector{});
```
mattsse, line 329: imo we should mandate an inspector argument, this is even weirder than having it by default in the evm type

PR #54, opened 2025-04-02
feat(tx): Add `FromTxWithEncoded` bound to `BlockExecutor` transaction

RomanHodulak:
## Motivation

Currently, the opstack use-case of `eth_simulate` on reth charges user L1 data cost due to the way tx envelope is derived for serialized transaction.

This is divergent from op-geth logic, where the L1 data cost is not charged.

The simulation fails for users with lower balance, where it should not fail.

## Solution

Allows for empty envelopes as part of https://github.com/paradigmxyz/reth/issues/15292

Empty envelopes will contain zero L1 data cost.

A `FromTxWithEncoded` trait bound is added to associated transaction type on `BlockExecutor`. This allows to create the tx envelope using `WithEncoded::new(Default::default(), tx)`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/evm/src/block/mod.rs
+++ b/crates/evm/src/block/mod.rs
@@ -1,6 +1,8 @@
 //! Block execution abstraction.
 
-use crate::{Database, Evm, EvmFactory, FromRecoveredTx, IntoTxEnv, RecoveredTx};
+use crate::{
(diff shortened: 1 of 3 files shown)


mattsse (commented):
I think this should work, but possible that we're missing some impls, so before we do this I need to check a few things in reth

PR #55, opened 2025-04-02
ci: rm precise pin

mattsse:


RomanHodulak (approved):
According to https://github.com/rust-lang/cc-rs/issues/1109

The issue is fixed at least in 1.0.103

PR #56, opened 2025-04-02
feat(tx): Implement `FromTxWithEncoded` and `FromRecoveredTx` for `EthereumTxEnvelope`

RomanHodulak:
Closes https://github.com/paradigmxyz/reth/issues/15454

## Motivation

Recently added trait bounds to transaction types include `FromTxWithEncoded` as seen in #54.

## Solution

* Implements `FromTxWithEncoded` for all variants of `TxEnv` that **ignores encoded bytes**
* Implements `FromTxWithEncoded` and `FromRecoveredTx` for `EthereumTxEnvelope`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/evm/src/tx.rs
+++ b/crates/evm/src/tx.rs
@@ -1,7 +1,8 @@
 //! Abstraction of an executable transaction.
 
 use alloy_consensus::{
-    transaction::Recovered, TxEip1559, TxEip2930, TxEip4844, TxEip7702, TxLegacy,
+    transaction::Recovered, EthereumTxEnvelope, TxEip1559, TxEip2930, TxEip4844, TxEip7702,
+    TxLegacy,
 };
 use alloy_eips::{eip2718::WithEncoded, Typed2718};
 use alloy_primitives::{Address, Bytes, TxKind};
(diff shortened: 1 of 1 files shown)


crates/evm/src/tx.rs
```diff
@@ -258,6 +289,36 @@ impl<T, TxEnv: FromTxWithEncoded<T>> IntoTxEnv<TxEnv> for &WithEncoded<Recovered
     }
 }
 
+impl<Eip4844, TxEnv> FromTxWithEncoded<EthereumTxEnvelope<Eip4844>> for TxEnv
+where
+    TxEnv: FromTxWithEncoded<TxLegacy>
+        + FromTxWithEncoded<TxEip2930>
+        + FromTxWithEncoded<Eip4844>
+        + FromTxWithEncoded<TxEip7702>
+        + FromTxWithEncoded<TxEip1559>,
```
mattsse, line 298: this should always hold, and also excludes the case `TxEip4844 for OpTxEnv`

crates/evm/src/tx.rs
```diff
@@ -258,6 +289,36 @@ impl<T, TxEnv: FromTxWithEncoded<T>> IntoTxEnv<TxEnv> for &WithEncoded<Recovered
...
+            EthereumTxEnvelope::Eip7702(tx) => TxEnv::from_encoded_tx(tx.tx(), caller, encoded),
+        }
+    }
+}
+
+impl<Eip4844, TxEnv> FromRecoveredTx<EthereumTxEnvelope<Eip4844>> for TxEnv
+where
+    Eip4844: Encodable2718 + RlpEcdsaEncodableTx,
+    TxEnv: FromTxWithEncoded<EthereumTxEnvelope<Eip4844>>,
+{
```
mattsse, line 315: I think this we can specialize for the Eth txenv because we don't want to encode anything here, which I think is fine because this is the actual envelope type.

and we can treat this similar to:

https://github.com/alloy-rs/evm/blob/b2a9720c4e804ce2c86533bbec3540a1c7d0d238/crates/evm/src/tx.rs#L270-L270

https://github.com/alloy-rs/evm/blob/b2a9720c4e804ce2c86533bbec3540a1c7d0d238/crates/evm/src/tx.rs#L314-L314

mattsse (changes requested):

mattsse (approved):

PR #62, opened 2025-04-08
feat: added method to get chain id

### LINKED ISSUE #61: [Feature] expose chain_id in evm trait

we currently provide access to the blockenv:

https://github.com/alloy-rs/evm/blob/65bdb46726a4cdb265f13e4ea663a57ecf0f8d6c/crates/evm/src/evm.rs#L47-L47

but we also want access to the chainid, which we can get from `self.cfg.chain_id`

## TODO
* add `fn chain_id` and return the chain id

this feature is useful for things like eth_simulate where we need to fill a tx

### Additional context

_No response_

Soubhik-10:
closes #61

### DIFF
--- a/crates/evm/src/env.rs
+++ b/crates/evm/src/env.rs
@@ -35,6 +35,11 @@ impl<Spec> EvmEnv<Spec> {
         &self.cfg_env
     }
 
+    /// Returns the chain ID of the environment.
+    pub const fn chainid(&self) -> u64 {
+        self.cfg_env.chain_id
+    }
+
     /// Returns the spec id of the chain
     pub const fn spec_id(&self) -> &Spec {
         &self.cfg_env.spec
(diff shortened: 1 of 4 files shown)


Soubhik-10:
@mattsse please check

mattsse (changes requested):
great, we can keep this one but we also want a trait function for this on:

https://github.com/alloy-rs/evm/blob/65bdb46726a4cdb265f13e4ea663a57ecf0f8d6c/crates/evm/src/evm.rs#L26-L26

similar to:

https://github.com/alloy-rs/evm/blob/65bdb46726a4cdb265f13e4ea663a57ecf0f8d6c/crates/evm/src/evm.rs#L46-L47

but `fn chain_id(&self) -> u64;`

Soubhik-10:
@mattsse please check..

mattsse (approved):

Issue #51, opened 2025-04-01
Integrate revm transact_system_call


rakita:
Revm added support for systems calls that skips validation, pre execution and most of the post execution phases simplifying usage and don't touch beneficiary/caller accounts. https://github.com/bluealloy/revm/pull/2350

alloy/evm function has a caller as an input, but all system calls use the same `SYSTEM_ADDRESS` revm requires only address and data as inputs.  
https://github.com/alloy-rs/evm/blob/b2a9720c4e804ce2c86533bbec3540a1c7d0d238/crates/evm/src/eth/mod.rs#L131

nadtech-hub:
I want to take on it

rakita:
go for it @nadtech-hub

nadtech-hub:
Sorry, I didn't notice that I had been unassigned. Why did you do it @rakita btw?

nadtech-hub:
EIP methods making system calls invoke set method if caller address is SYSTEM_ADDRESS. How about get methods?

rakita:
Hey @nadtech-hub I unsigned you as this task is sensitive and it would be best done by one of the team member.

I apologize for the confusion.

nadtech-hub:
Thanks for clarifying @rakita 
If you could also review my PR. This task didn't seem challenging.

PR #68, opened 2025-04-15
feat: transact_system_call refactoring

nadtech-hub:
## Motivation

This should close [#51](https://github.com/alloy-rs/evm/issues/51#event-17248333584)

## Solution

Set system address as default caller's address. Delegate system calls to low-level revm 

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/evm/src/eth/mod.rs
+++ b/crates/evm/src/eth/mod.rs
@@ -149,8 +149,16 @@ where
         self.inspect = enabled;
     }
 
+    fn precompiles(&self) -> &Self::Precompiles {
+        &self.inner.precompiles
+    }
+
     fn precompiles_mut(&mut self) -> &mut Self::Precompiles {
         &mut self.inner.precompiles
+    }
+
+    fn inspector(&self) -> &Self::Inspector {
+        &self.inner.inspector
     }
 
     fn inspector_mut(&mut self) -> &mut Self::Inspector {
--- a/crates/evm/src/eth/mod.rs
+++ b/crates/evm/src/eth/mod.rs
@@ -1,8 +1,7 @@
 //! Ethereum EVM implementation.
 
 use crate::{env::EvmEnv, evm::EvmFactory, precompiles::PrecompilesMap, Database, Evm};
-use alloc::vec::Vec;
-use alloy_primitives::{Address, Bytes, TxKind, U256};
+use alloy_primitives::{Address, Bytes};
 use core::{
     fmt::Debug,
     ops::{Deref, DerefMut},
@@ -15,7 +14,7 @@ use revm::{
     interpreter::{interpreter::EthInterpreter, InterpreterResult},
     precompile::{PrecompileSpecId, Precompiles},
     primitives::hardfork::SpecId,
-    Context, ExecuteEvm, InspectEvm, Inspector, MainBuilder, MainContext,
+    Context, ExecuteEvm, InspectEvm, Inspector, MainBuilder, MainContext, SystemCallEvm,
 };
 
 mod block;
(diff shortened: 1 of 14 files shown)


nadtech-hub:
@rakira all refactored EIP calls invoke set method because caller address is SYSTEM_ADDRESS. How about get methods?

klkvr (commented):
I think the idea in #51 is that we should use the newly introduced methods from https://github.com/bluealloy/revm/pull/2350 (`transact_system_call` from `SystemCallEvm` trait)

nadtech-hub:
> I think the idea in #51 is that we should use the newly introduced methods from [bluealloy/revm#2350](https://github.com/bluealloy/revm/pull/2350) (`transact_system_call` from `SystemCallEvm` trait)

I referenced revm's system call implementation while refactoring this evm's methods. Should've I added new methods instead of refactoring existing?

klkvr:
the EVMs in this repo wrap revm EVMs, we need to use the newly added methods on revm EVMs directly

this would likely also require bumping revm

nadtech-hub:
therefore title of issue says integrate. I will remake it

crates/evm/src/eth/mod.rs
```diff
@@ -187,7 +145,7 @@ where
         // We're doing this state cleanup to make sure that changeset only includes the changed
         // contract storage.
         if let Ok(res) = &mut res {
-            res.state.retain(|addr, _| *addr == contract);
+            res.state.retain(|addr, _| *addr == callee);
```
nadtech-hub, line 148: I doubt about retain fn call here. I will try to link this evm wtih reth for testing

rakita: This part can be removed, lines 145-147. Revm now returns only system contract changes

crates/op-evm/src/lib.rs
```diff
@@ -186,7 +135,7 @@ where
         // We're doing this state cleanup to make sure that changeset only includes the changed
         // contract storage.
         if let Ok(res) = &mut res {
```
rakita, line 137: Same here, this parts can be removed lines 137-139

rakita (approved):
lgtm, pending nits

nadtech-hub:
@klkvr @rakita @mattsse how to configure fmt properly on my machine?  I didn't see the issue which caused job failure.

crates/evm/src/eth/mod.rs
```diff
@@ -134,65 +133,10 @@ where
 
     fn transact_system_call(
         &mut self,
-        caller: Address,
-        contract: Address,
+        callee: Address,
         data: Bytes,
```
mattsse, line 137: I prefer having caller first, then data, which is inverse to what the revm trait expects, is fine but ideally we're consistent here but I find address,data more intuitive

mattsse (commented):
cool, this does get rid of a bunch of hacky workaround code.

supporting this natively in revm makes sense.

I haven't looked into the revm trait and impl yet

crates/evm/src/eth/mod.rs
rakita: It does makes sense to do the caller first, will see to add that change before releasing

nadtech-hub: But caller is always SYSTEM_ADRESS as per issue's description.

nadtech-hub: That's part I understood that @mattsse wants to see that change in revm.

mattsse (commented):
superseded by #121

PR #71, opened 2025-04-17
feat: add SpecPrecompiles

### LINKED ISSUE #70: [Feature] Add a way to replace the precompileprovider

for things like `eth_simulateV1` that support moving precompiles (for some reason), or use cases like replacing precompiles.

we should add a way to modify those via the evm(factory) interface

### Additional context

_No response_

mattsse:
closes #70

instead of making possible to configure a custom precompiles provider on the factory level, we should instead make it possible to modify the precompiles via the evm on demand.

but we can also solve this with another constructor fn on EvmFactory, however, then we still need to use one specific PrecompileProvider type.

since this turned out to be a bit tricky with GATs, we can introduce our own Precompiles type that supports the usecases we want: hotswapping function pointers for example.

and we can offload all custom functionality directly into the precompile fn pointer instead of doing this: https://github.com/paradigmxyz/reth/pull/15536

### DIFF
--- a/crates/evm/src/precompiles.rs
+++ b/crates/evm/src/precompiles.rs
@@ -241,16 +241,6 @@ impl From<PrecompileFn> for DynPrecompile {
     }
 }
 
-impl DynPrecompile {
-    /// Wraps this precompile with a custom implementation.
-    pub fn wrap<P: Precompile + Send + Sync + 'static>(
-        self,
-        wrapper: impl FnOnce(Self) -> P,
-    ) -> Self {
-        Self(Arc::new(wrapper(self)))
-    }
-}
-
 /// A mutable representation of precompiles that allows for runtime modification.
 ///
 /// This structure stores dynamic precompiles that can be modified at runtime,
--- a/crates/evm/src/precompiles.rs
+++ b/crates/evm/src/precompiles.rs
@@ -1,7 +1,9 @@
 //! Helpers for dealing with Precompiles.
 
 use alloc::borrow::Cow;
+use alloy_primitives::Address;
 use revm::{handler::EthPrecompiles, precompile::Precompiles, primitives::hardfork::SpecId};
+use revm::precompile::PrecompileFn;
 
 /// A set of Precompiles according to a spec.
 #[derive(Debug, Clone)]
@@ -14,6 +16,7 @@ pub struct SpecPrecompiles<Spec> {
     spec: Spec,
 }
 
+
 impl<Spec> SpecPrecompiles<Spec> {
     /// Creates the [`SpecPrecompiles`] from a static reference.
     pub const fn from_static(precompiles: &'static Precompiles, spec: Spec) -> Self {
@@ -23,6 +26,27 @@ impl<Spec> SpecPrecompiles<Spec> {
     /// Creates a new set of precompiles for that spec.
     pub fn new(precompiles: Cow<'static, Precompiles>, spec: Spec) -> Self {
         Self { precompiles, spec }
+    }
+    
+    /// Returns the configured precompiles
+    pub fn precompiles(&self) -> &Precompiles {
+        &self.precompiles
+    }
+    
+    /// Returns mutable access to the precompiles.
+    pub fn precompiles_mut(&mut self) -> &mut Precompiles {
+        self.precompiles.to_mut()
+    }
+    
+    /// Returns the configured spec.
+    pub const fn spec(&self) -> &Spec {
+        &self.spec
+    }
+    
+    pub fn map_precompile<F>(&mut self, address: &Address, f: F) where F: Fn(PrecompileFn) -> PrecompileFn {
+        if let Some(precompile) = self.precompiles_mut().get(address) {
+            
+        }
     }
 }
 
--- a/crates/evm/src/precompiles.rs
+++ b/crates/evm/src/precompiles.rs
@@ -26,7 +26,7 @@ pub enum PrecompilesMap {
 }
 
 impl PrecompilesMap {
-    /// Creates the [`SpecPrecompiles`] from a static reference.
+    /// Creates the [`PrecompilesMap`] from a static reference.
     pub fn from_static(precompiles: &'static Precompiles) -> Self {
         Self::new(Cow::Borrowed(precompiles))
     }
--- a/crates/evm/src/precompiles.rs
+++ b/crates/evm/src/precompiles.rs
@@ -31,34 +31,67 @@ impl<Spec> SpecPrecompiles<Spec> {
         Self { precompiles: PrecompilesMap::Builtin(precompiles), spec }
     }
 
-    /// Returns the configured precompiles
-    pub fn precompiles(&self) -> &Precompiles {
+    /// Returns the configured precompiles as a PrecompilesMut
+    ///
+    /// This converts static Precompiles to a dynamic representation
+    /// that can be modified.
+    pub fn precompiles(&self) -> PrecompilesMut {
         match &self.precompiles {
-            PrecompilesMap::Builtin(cow) => cow,
-            PrecompilesMap::Dynamic(_) => {
-                todo!()
+            PrecompilesMap::Builtin(cow) => {
+                let mut dynamic = PrecompilesMut::default();
+
+                let precompiles = match cow {
+                    Cow::Borrowed(static_ref) => *static_ref,
+                    Cow::Owned(ref owned) => owned,
+                };
+
+                // convert all static precompiles to dynamic ones
+                for (addr, precompile_fn) in precompiles.inner() {
+                    let dyn_precompile: DynPrecompile = (*precompile_fn).into();
+                    dynamic.inner.insert(*addr, dyn_precompile);
+                    dynamic.addresses.insert(*addr);
+                }
+
+                dynamic
             }
+            PrecompilesMap::Dynamic(dynamic) => dynamic.clone(),
         }
     }
 
-    /// Returns mutable access to the precompiles.
-    pub fn precompiles_mut(&mut self) -> &mut Precompiles {
-        match &mut self.precompiles {
+    /// Returns mutable access to the precompiles as a PrecompilesMut.
+    ///
+    /// If the current representation is static, this will convert it to a dynamic
+    /// representation and store it for future use. If it's already dynamic, this
+    /// will return a new mutable instance with the same data.
+    ///
+    /// This does not return a direct reference to the internal precompiles, but instead
+    /// a new instance that can be modified and then later applied back
+    pub fn precompiles_mut(&mut self) -> PrecompilesMut {
+        match &self.precompiles {
             PrecompilesMap::Builtin(cow) => {
-                // Ensure we have a mutable Cow by cloning the borrowed data if needed
-                if let Cow::Borrowed(static_ref) = cow {
-                    *cow = Cow::Owned((*static_ref).clone());
+                let mut dynamic = PrecompilesMut::default();
+
+                let precompiles = match cow {
+                    Cow::Borrowed(static_ref) => *static_ref,
+                    Cow::Owned(ref owned) => owned,
+                };
+
+                // convert all static precompiles to dynamic ones
+                for (addr, precompile_fn) in precompiles.inner() {
+                    let dyn_precompile: DynPrecompile = (*precompile_fn).into();
+                    dynamic.inner.insert(*addr, dyn_precompile);
+                    dynamic.addresses.insert(*addr);
                 }
-                // Now we definitely have an owned value
-                match cow {
-                    Cow::Owned(ref mut owned) => owned,
-                    Cow::Borrowed(_) => unreachable!("We just ensured we have an owned value"),
-                }
+
+                dynamic
             }
-            PrecompilesMap::Dynamic(_) => {
-                todo!()
-            }
+            PrecompilesMap::Dynamic(dynamic) => dynamic.clone(),
         }
+    }
+
+    /// Sets the precompiles to the given dynamic representation.
+    pub fn set_precompiles(&mut self, precompiles: PrecompilesMut) {
+        self.precompiles = PrecompilesMap::Dynamic(precompiles);
     }
 
     /// Returns the configured spec.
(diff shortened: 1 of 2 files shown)


crates/evm/src/precompiles.rs
```diff
@@ -0,0 +1,331 @@
...
+
+    /// Returns mutable access to the precompiles as a PrecompilesMut.
+    ///
+    /// If the current representation is static, this will convert it to a dynamic
+    /// representation and store it for future use. If it's already dynamic, this
+    /// will return a new mutable instance with the same data.
+    ///
+    /// This does not return a direct reference to the internal precompiles, but instead
+    /// a new instance that can be modified and then later applied back
+    pub fn precompiles_mut(&mut self) -> PrecompilesMut {
```
mattsse, line 69: hmm, it feels a bit strange that this returns a clone.

it looks like this doesn't actually mutate the precompiles in place?

crates/evm/src/precompiles.rs
```diff
@@ -0,0 +1,331 @@
...
+        Self(Arc::new(f))
+    }
+}
+
+/// A mutable representation of precompiles that allows for runtime modification.
+///
+/// This structure stores dynamic precompiles that can be modified at runtime,
+/// unlike the static `Precompiles` struct from revm.
+#[derive(Clone, Default)]
+pub struct PrecompilesMut {
```
mattsse, line 239: should we rename this to `DynPrecompiles`?

crates/evm/src/precompiles.rs
```diff
@@ -0,0 +1,331 @@
...
+    /// Precompiles
+    pub inner: HashMap<Address, DynPrecompile>,
+    /// Addresses of precompile
+    pub addresses: HashSet<Address>,
```
mattsse, line 243: I'd prefer if we keep this private

crates/evm/src/precompiles.rs
```diff
@@ -0,0 +1,331 @@
...
+    fn from(value: EthPrecompiles) -> Self {
+        Self::from_static(value.precompiles, value.spec)
+    }
+}
+
+#[derive(Clone)]
+enum PrecompilesMap {
+    Builtin(Cow<'static, Precompiles>),
+    Dynamic(PrecompilesMut),
+}
```
mattsse, line 194: here we can add 

```rust
fn as_dyn_precompiles(&mut self) -> &mut PrecompilesMut {
  // if builtin convert into Dyn
} 

```

I believe this would simplify a lot of code in the functions above

mattsse (commented):
the apply/map should provide most of the features we need.

what we also need is a `PrecompileProvider` impl for that, that mirrors the EthPrecompiles impl, because in order to exposes these features in the Op and Ethevm we'd need to the precompile type they operate on

ptal as well @klkvr

crates/evm/src/precompiles.rs
fgimenez: indeed, this was due to some issues with the dual precompiles we had. i have refactored a bit, now the transformation static -> dynamic is done in the constructor and internally we always manage dyn precompiles, this makes `precompiles` and `precompiles_mut` getters behave as they should, ptal.

crates/evm/src/precompiles.rs
fgimenez: not needed with the current approach, we do this once in the constructor. if we want to include static precompiles at runtime we could extract this logic to a separate method, wdyt?

crates/evm/src/precompiles.rs
fgimenez: much better, done

crates/evm/src/precompiles.rs
fgimenez: makes sense, done

crates/evm/src/precompiles.rs
```diff
@@ -0,0 +1,254 @@
...
+    primitives::hardfork::SpecId,
+};
+
+/// A set of Precompiles according to a spec.
+#[derive(Debug, Clone)]
+pub struct SpecPrecompiles<Spec> {
+    /// The configured precompiles.
+    precompiles: DynPrecompiles,
+    /// The spec these precompiles belong to.
+    spec: Spec,
```
klkvr, line 22: I don't think we need `spec` here at all? given that precompiles set is static and corresponds to a single EVM with fixed spec

crates/evm/src/precompiles.rs
```diff
@@ -0,0 +1,254 @@
...
+
+        // Convert all static precompiles to dynamic ones
+        for (addr, precompile_fn) in static_precompiles.inner() {
+            let dyn_precompile: DynPrecompile = (*precompile_fn).into();
+            dynamic.inner.insert(*addr, dyn_precompile);
+            dynamic.addresses.insert(*addr);
+        }
+
+        Self { precompiles: dynamic, spec }
+    }
```
klkvr: wouldn't it be beneficial to store static precompiles by default? and convert them to dynamic on first mutating call

crates/evm/src/precompiles.rs
```diff
@@ -0,0 +1,254 @@
...
+        let precompiles = self.precompiles_mut();
+
+        // Get the current precompile at the address
+        if let Some(dyn_precompile) = precompiles.inner.get(address).cloned() {
+            // Apply the transformation function
+            let transformed = f(dyn_precompile);
+
+            // Update the precompile at the address
+            precompiles.inner.insert(*address, transformed);
+        }
```
klkvr, line 65:
```suggestion
        // Get the current precompile at the address
        if let Some(dyn_precompile) = precompiles.inner.remove(address) {
            // Apply the transformation function
            let transformed = f(dyn_precompile);

            // Update the precompile at the address
            precompiles.inner.insert(*address, transformed);
        }
```

will this work?

fgimenez:
> what we also need is a `PrecompileProvider` impl for that, that mirrors the EthPrecompiles impl, because in order to exposes these features in the Op and Ethevm we'd need to the precompile type they operate on

sure, done here https://github.com/alloy-rs/evm/pull/71/commits/1d37695f37b3363772e7dc5d7cd0700c55751a47

crates/evm/src/precompiles.rs
fgimenez: finally added a separate `into_dyn_precompiles` with this functionality, it is used also in `PrecompileProvider` impl

crates/evm/src/precompiles.rs
```diff
@@ -0,0 +1,326 @@
...
+        if spec == self.spec {
+            return false;
+        }
+
+        self.precompiles = into_dyn_precompiles(Cow::Borrowed(Precompiles::new(
+            PrecompileSpecId::from_spec_id(spec.clone().into()),
+        )));
+        self.spec = spec;
+        true
+    }
```
klkvr, line 156: this will only work for eth but ideally we want to be able to plug this type into op as well

I think what we could do is have conversion `From<EthPrecompiles` and `From<OpPrecompiles>` an then in eth/op evm factory we'd simply instantiate the corresponding revm's precompiles based on spec and just convert them into this type when setting up the actual EVM.

I would really prefer this to not have spec generic as this is supposed to be a simple set of fn pointers

crates/evm/src/precompiles.rs
fgimenez: makes sense, removed

crates/evm/src/precompiles.rs
fgimenez: indeed, this was the initial approach, i've taken it back, ptal

crates/evm/src/precompiles.rs
fgimenez: makes sense, spec generic is removed. we already have the `From<EthPrecompiles>`, for op we need to access inner.precompiles, I've added a todo.

crates/evm/src/precompiles.rs
```diff
@@ -0,0 +1,422 @@
...
+
+            for (addr, precompile_fn) in static_precompiles.inner() {
+                let dyn_precompile: DynPrecompile = (*precompile_fn).into();
+                dynamic.inner.insert(*addr, dyn_precompile);
+                dynamic.addresses.insert(*addr);
+            }
+
+            self.precompiles = PrecompilesMap::Dynamic(dynamic);
+        }
+    }
```
klkvr, line 129: nit: let's do `pub fn as_dynamic_mut(&mut self) -> &mut DynPrecompiles`

I think it'd be useful generally and will also allow to simplify other methods a bit

crates/evm/src/precompiles.rs
```diff
@@ -0,0 +1,422 @@
...
+}
+
+impl PrecompilesMap {
+    /// Returns all precompile addresses.
+    fn addresses(&self) -> Vec<Address> {
+        match self {
+            Self::Builtin(precompiles) => precompiles.addresses().copied().collect(),
+            Self::Dynamic(dyn_precompiles) => dyn_precompiles.addresses.iter().copied().collect(),
+        }
+    }
```
klkvr, line 164: I'd prefer this to return a slice or iterator over `&Address`, let's leave up to the callsite whether we want to allocate

crates/evm/src/precompiles.rs
```diff
@@ -0,0 +1,422 @@
...
+
+impl PartialEq for PrecompilesMap {
+    fn eq(&self, other: &Self) -> bool {
+        match (self, other) {
+            (Self::Builtin(_), Self::Builtin(_)) => true,
+            (Self::Dynamic(a), Self::Dynamic(b)) => a.addresses == b.addresses,
+            _ => false,
+        }
+    }
+}
```
klkvr, line 175: why do we need this?

crates/evm/src/precompiles.rs
```diff
@@ -0,0 +1,422 @@
...
+                }
+            }
+            PrecompilesMap::Dynamic(dyn_precompiles) => {
+                if let Some(precompile) = dyn_precompiles.inner.get(address) {
+                    precompile.call(&inputs.input, gas_limit)
+                } else {
+                    return Ok(Some(result)); // Should never happen due to has_precompile check
+                }
+            }
+        };
```
klkvr, line 263: could we have and use a `fn get(&self, address: &Address) -> Option<impl Precompile>` API here?

crates/evm/src/precompiles.rs
```diff
@@ -0,0 +1,422 @@
...
+
+        if let PrecompilesMap::Dynamic(ref mut dyn_precompiles) = self.precompiles_mut() {
+            // apply the transformation to each precompile
+            let mut new_map = HashMap::new();
+            for (addr, precompile) in &dyn_precompiles.inner {
+                let transformed = f(addr, precompile.clone());
+                new_map.insert(*addr, transformed);
+            }
+
+            dyn_precompiles.inner = new_map;
```
klkvr, line 78: we could also avoid cloning here, though clones are cheap rn so it doesn't really matter

btw how should stateful (e.g cached precompiles look like in this context)? should we add a way to mutate the state of precompile or would they use Mutex internally?

previous revm had this https://github.com/bluealloy/revm/blob/bafe64b64a67bcd4db62729ba29df94caa38c53b/crates/revm/src/context/context_precompiles.rs#L10-L21 which allowed to `Box` a precompile and invoke it mutably

klkvr (changes requested):
I guess we can just merge `PrecompilesMap` with `SpecPrecompiles` now?

crates/evm/src/precompiles.rs
mattsse: right we can operate on impl now

klkvr: was curious if we can make this work with `either::Either` and simplify a bit, seems good 54933dd

crates/evm/src/precompiles.rs
fgimenez: proposed revm change here https://github.com/bluealloy/revm/pull/2444 added link to the TODO comment

crates/evm/src/precompiles.rs
fgimenez: ok done here https://github.com/alloy-rs/evm/pull/71/commits/4cc732c78c150f4c7c0ebfb5e9950330cf9f530b

crates/evm/src/precompiles.rs
jenpaff: @fgimenez is this this https://github.com/bluealloy/revm/pull/2444 blocking the PR?

crates/evm/src/precompiles.rs
fgimenez: > we could also avoid cloning here, though clones are cheap rn so it doesn't really matter

done here https://github.com/alloy-rs/evm/pull/71/commits/ee96b957cefaf9770885afea760ac54b7494eab2

> btw how should stateful (e.g cached) precompiles look like in this context? should we add a way to mutate the state of precompile or would they use Mutex internally?

we aim to have something like:
```rust
CachedProcompile {
  precompile : DynPrecompile,
  cache: Arc<cache>
}
```
and
```rust
impl Precompile for CachedPrecompile
```
then we can call this like `evm.map_precompiles(|_, precompile| CachedPrecompile::wrap(precompile, cache.clone()).into())` from the engine

we are planning to use `mini-moka` for the `cache` field, which is thread safe, so i think we should be ok, no need to add any other ways to access and mutate state, should be handled internally.

crates/evm/src/precompiles.rs
fgimenez: don't think so, we can go ahead on eth without it and implement the changes for op once it is released

crates/evm/src/precompiles.rs
```diff
@@ -0,0 +1,428 @@
...
+
+    /// Returns the configured precompiles as a read-only reference.
+    pub fn precompiles(&self) -> &Self {
+        self
+    }
+
+    /// Returns mutable access to the precompiles as a DynPrecompiles.
+    pub fn precompiles_mut(&mut self) -> &mut Self {
+        self
+    }
```
klkvr, line 47: those don't seem useful?

fgimenez: yep, removed

klkvr (approved):
lgtm, but I guess we'd still need op-revm fix for this to be integrate into evm trait?

fgimenez:
> lgtm, but I guess we'd still need op-revm fix for this to be integrate into evm trait?

thx, could we do a no-op impl for the new method in op until the release is out? I'll prepare a followup with the remaining changes, we can discuss there or hold it until new op-revm is released, wdyt?

klkvr:
I think our latest idea was to do the following

```
trait Evm {
    type Precompiles;
    
    fn precompiles_mut(&mut self) -> &mut Self::Precompiles;
}
```

and on reth side we'd just always require `Evm: Evm<Precompiles = PrecompilesMap>`. So this would only work if all EVMs used inside of reth would expose this

I'd be OK with us operating on patches for now to iterate faster here

fgimenez:
> I think our latest idea was to do the following
> 
> ```
> trait Evm {
>     type Precompiles;
>     
>     fn precompiles_mut(&mut self) -> &mut Self::Precompiles;
> }
> ```
> 
> and on reth side we'd just always require `Evm: Evm<Precompiles = PrecompilesMap>`. So this would only work if all EVMs used inside of reth would expose this
> 
> I'd be OK with us operating on patches for now to iterate faster here

makes sense thx!

PR #73, opened 2025-04-24
feat: include Precompiles associated type in Evm trait

fgimenez:
towards #70

## Motivation

After #71, which allowed modifying precompiles at runtime, we need to make the functionality accessible to evm consumers.

## Solution

Adds a `Precompiles` associated type and a `precompiles_mut` method to the `Evm` trait.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/evm/src/eth/mod.rs
+++ b/crates/evm/src/eth/mod.rs
@@ -272,30 +272,50 @@ mod tests {
 
     #[test]
     fn test_precompiles_with_correct_spec() {
-        // Test with different spec IDs to ensure precompiles are correctly initialized
+        // create tests where precompile should be available for later specs but not earlier ones
         let specs_to_test = [
-            (SpecId::FRONTIER, address!("0x0000000000000000000000000000000000000001")), /* ECRECOVER only in Frontier */
-            (SpecId::BYZANTIUM, address!("0x0000000000000000000000000000000000000005")), /* MODEXP added in Byzantium */
-            (SpecId::ISTANBUL, address!("0x0000000000000000000000000000000000000009")), /* BLAKE2F added in Istanbul */
+            // MODEXP (0x05) was added in Byzantium, should not exist in Frontier
+            (
+                address!("0x0000000000000000000000000000000000000005"),
+                SpecId::FRONTIER,  // Early spec - should NOT have this precompile
+                SpecId::BYZANTIUM, // Later spec - should have this precompile
+                "MODEXP",
+            ),
+            // BLAKE2F (0x09) was added in Istanbul, should not exist in Byzantium
+            (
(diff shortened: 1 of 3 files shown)


crates/evm/src/evm.rs
```diff
@@ -42,6 +42,8 @@ pub trait Evm {
     /// Identifier of the EVM specification. EVM is expected to use this identifier to determine
     /// which features are enabled.
     type Spec: Debug + Copy + Send + Sync + 'static;
+    /// Precompiles used by the EVM.
+    type Precompiles;
```
mattsse, line 46: should we add some trait bounds for this so that we can access the functionality that we want from reth pov without adding there?

mattsse, line 46: I like this, this shouldn't be problematic because we assume this type doesn't need any generics that aren't already part of this trait, like Spec.

mattsse (commented):
lgtm,

only have one q re default trait bounds @klkvr

klkvr:
should we also expose it on `EvmFactory` as we do with other ATs?

Right now it's impossible to bound `Factory::Evm::Precompiles` because of GAT

fgimenez:
> should we also expose it on `EvmFactory` as we do with other ATs?

good point, done here https://github.com/alloy-rs/evm/pull/73/commits/e6129340de1613f700dbae774ad73aa56095285c, in the Eth impl i have already set the precompiles to `PrecompileMap`, no changes for op, let me know wdyt

crates/evm/src/eth/mod.rs
```diff
@@ -214,20 +219,24 @@ where
...
         EthEvm {
             inner: Context::mainnet()
                 .with_block(input.block_env)
                 .with_cfg(input.cfg_env)
                 .with_db(db)
-                .build_mainnet_with_inspector(NoOpInspector {}),
+                .build_mainnet_with_inspector(NoOpInspector {})
+                .with_precompiles(PrecompilesMap::from_static(
+                    EthPrecompiles::default().precompiles,
+                )),
```
klkvr, line 239: I think this is wrong and will always inject precompiles from latest spec

we should do something like `PrecompilesMap::from_static(Precompiles::new(PrecompileSpecId::from_spec_id(spec)))`

when using `EthPrecompiles` revm is setting correct spec at runtime but we want to have it fixed as this is how our API is designed

klkvr: could also add a basic test for this

fgimenez: great catch! done and added a test, ptal

klkvr (approved):
lgtm

PR #77, opened 2025-05-06
chore: re-export revm & op-revm

n0b0dyCN:
## Motivation

Re-export revm and op-revm for easier version control.

## Solution

Re-export them in `crates/evm/src/lib.rs`.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/evm/src/lib.rs
+++ b/crates/evm/src/lib.rs
@@ -23,6 +23,6 @@ pub use tx::*;
 pub mod precompiles;
 
 // re-export revm and op-revm
-pub use revm;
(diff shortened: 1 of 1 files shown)


mattsse (approved):
I think that makes sense

n0b0dyCN:
> I think that makes sense

Thanks for your reply. Since I'm also using revm to build something else, I think it is good to also export some feature flags in revm, e.g., `optional_balance_check`. If you think this is ok, I can continue to add more feature control in this PR.

PR #81, opened 2025-05-08
feat: expose Inspector on Evm

### LINKED ISSUE #76: [Feature] Expose inspector on Evm

for https://github.com/paradigmxyz/reth/issues/16045 to be feasible we need access to inspector on Evm trait.

We need to add `Inspector` AT and expose `inspector_mut` on https://github.com/alloy-rs/evm/blob/f3ea0bc563c4259aa9c02288e6ec70fc438c926d/crates/evm/src/evm.rs#L26

And also lock the AT to configured generic here https://github.com/alloy-rs/evm/blob/f3ea0bc563c4259aa9c02288e6ec70fc438c926d/crates/evm/src/evm.rs#L144-L152

### Additional context

_No response_

fgimenez:
## Motivation

Closes #76 

Adds AT and mutable getter for inspector to `Evm` trait. Also locks the AT in `EvmFactory to the configured generic.`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/evm/src/eth/mod.rs
+++ b/crates/evm/src/eth/mod.rs
@@ -79,11 +79,6 @@ impl<DB: Database, I, PRECOMPILE> EthEvm<DB, I, PRECOMPILE> {
     pub fn ctx_mut(&mut self) -> &mut EthEvmContext<DB> {
         &mut self.inner.ctx
     }
-
-    /// Provides a mutable reference to the EVM inspector.
-    pub fn inspector_mut(&mut self) -> &mut I {
-        &mut self.inner.inspector
-    }
 }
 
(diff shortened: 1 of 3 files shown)


mattsse (changes requested):
simple enough, this makes sense because revm requires an Inspector type on evm level anyway, so we can do this.

can we remove the native pub funcions:

https://github.com/alloy-rs/evm/blob/480dc5e96333eb77fe035f3d1873cb74ac2b6825/crates/evm/src/eth/mod.rs#L83-L86

so that this doesn't clash

fgimenez:
> can we remove the native pub funcions:

good catch, done

mattsse (approved):
lgtm, pending @klkvr

klkvr (approved):
nice

PR #82, opened 2025-05-08
feat(eip7702): Delegate signer recovery to `alloy-consensus::crypto`

### LINKED ISSUE #75: [Feature] Use alloy_consensus crypto for 7702 recovery

https://github.com/alloy-rs/alloy/pull/2352 added support for secp recovery in alloy.

we can use this when we populate the revm txenv:

https://github.com/alloy-rs/evm/blob/71d62c642603bfcd2f71ee61ae9d17f16309c10f/crates/evm/src/tx.rs#L208-L208

but for this we need a new revm release first until then this is blocked

### Additional context

_No response_

RomanHodulak:
Closes #75 

Leverages faster `secp256k1` crate based algorithm if `std` feature flag is on

## Motivation

The `alloy` have a new sender recovery algorithm using `secp256k1` crate. It was found in benchmarks that this crate's algorithm is significantly faster than the existing `k256`, but requires `std` making it usable only on platforms that allow it.

## Solution

Recover authority list of EIP-7702 transaction using the `alloy_consensus::crypto` module to recover the signer.

## PR Checklist

- [X] Added Tests
- [X] Added Documentation
- [X] Breaking changes

### DIFF
--- a/crates/evm/Cargo.toml
+++ b/crates/evm/Cargo.toml
@@ -14,9 +14,9 @@ repository.workspace = true
 workspace = true
 
 [dependencies]
+alloy-consensus = { workspace = true, features = ["k256"] }
 alloy-primitives.workspace = true
 alloy-sol-types.workspace = true
-alloy-consensus.workspace = true
 alloy-eips.workspace = true
 alloy-hardforks.workspace = true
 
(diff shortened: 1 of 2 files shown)


mattsse (approved):
simple enough

lgtm!

PR #84, opened 2025-05-09
impl Evm for Either

nadtech-hub:
## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


nadtech-hub:
I've been struggling to resolve ownership issue in block fn :)

crates/evm/src/precompiles.rs
nadtech-hub, line 1: I pushed it by accident

mattsse (approved):

klkvr (approved):
thank you! it seems that impl for `EvmFactory` is not feasible rn because of some GAT restrictions so we'd need to tackle this separately

PR #86, opened 2025-05-12
fix: `r.as_ref()` the trait `AsRef<[_; 0]>` is not implemented for `[u8]`

zerosnacks:
## Motivation

Whilst bumping to `alloy-evm` v0.7.1 I encountered this issue

This only occurs when `alloy-evm` is integrated (in this case Foundry); it is not raised when I clone and compile `alloy-evm` from source

Screenshot from 2025-05-12 14-49-46

## Solution

I think this code is equivalent and resolves the compiler error downstream for Foundry

cc @mattsse / @klkvr 

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/evm/src/precompiles.rs
+++ b/crates/evm/src/precompiles.rs
@@ -1,7 +1,5 @@
 //! Helpers for dealing with Precompiles.
 
-use core::cell::Ref;
-
 use alloc::{borrow::Cow, boxed::Box, string::String, sync::Arc};
 use alloy_consensus::transaction::Either;
 use alloy_primitives::{
--- a/crates/evm/src/precompiles.rs
+++ b/crates/evm/src/precompiles.rs
@@ -190,11 +190,16 @@ impl<CTX: ContextTr> PrecompileProvider<CTX> for PrecompilesMap {
         };
 
         // Execute the precompile
-        let mut _guard: Option<Ref<'_, [u8]>> = None;
-        let input_bytes: &[u8] = match &inputs.input {
+        let r;
+        let input_bytes = match &inputs.input {
             CallInput::SharedBuffer(range) => {
-                _guard = context.local().shared_memory_buffer_slice(range.clone());
-                _guard.as_deref().unwrap_or(&[])
+                match context.local().shared_memory_buffer_slice(range.clone()) {
+                    Some(slice) => {
+                        r = slice;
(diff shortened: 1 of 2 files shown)


crates/evm/src/precompiles.rs
```diff
@@ -188,16 +190,11 @@ impl<CTX: ContextTr> PrecompileProvider<CTX> for PrecompilesMap {
...
             CallInput::SharedBuffer(range) => {
-                match context.local().shared_memory_buffer_slice(range.clone()) {
-                    Some(slice) => {
-                        r = slice;
-                        r.as_ref()
-                    }
-                    None => &[],
-                }
+                _guard = context.local().shared_memory_buffer_slice(range.clone());
+                _guard.as_deref().unwrap_or(&[])
```
mattsse, line 197: ngl, this looks hideous :D

crates/evm/src/precompiles.rs
```diff
@@ -188,16 +190,11 @@ impl<CTX: ContextTr> PrecompileProvider<CTX> for PrecompilesMap {
...
         // Execute the precompile
-        let r;
-        let input_bytes = match &inputs.input {
+        let mut _guard: Option<Ref<'_, [u8]>> = None;
+        let input_bytes: &[u8] = match &inputs.input {
             CallInput::SharedBuffer(range) => {
-                match context.local().shared_memory_buffer_slice(range.clone()) {
-                    Some(slice) => {
-                        r = slice;
-                        r.as_ref()
```
mattsse, line 197: can we do `&*r` here instead, I believe this is how we solved htis in revm-inspectors

https://github.com/paradigmxyz/revm-inspectors/commit/5bf3f108ea8cea22c4c905eaac39e1de2bd7a926

mattsse (changes requested):

crates/evm/src/precompiles.rs
zerosnacks: yeah agreed

crates/evm/src/precompiles.rs
zerosnacks: nice, this works :+1:

zerosnacks: much better :smile:

PR #88, opened 2025-05-14
feat: extend Evm::Spec bounds with Hash and PartialEq

fgimenez:
## Motivation

We need to include EVM's `Spec` AT  in reth's precompile cache key for https://github.com/paradigmxyz/reth/issues/16207 this PR adds the missing bounds required

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/evm/src/evm.rs
+++ b/crates/evm/src/evm.rs
@@ -166,7 +166,7 @@ pub trait EvmFactory {
     /// Halt reason. See [`Evm::HaltReason`].
     type HaltReason: HaltReasonTr + Send + Sync + 'static;
     /// The EVM specification identifier, see [`Evm::Spec`].
-    type Spec: Debug + Copy + Send + Sync + 'static;
+    type Spec: Debug + Copy + Hash + PartialEq + Send + Sync + 'static;
     /// Precompiles used by the EVM.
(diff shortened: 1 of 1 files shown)


crates/evm/src/evm.rs
```diff
@@ -41,7 +41,7 @@ pub trait Evm {
     type HaltReason: HaltReasonTr + Send + Sync + 'static;
     /// Identifier of the EVM specification. EVM is expected to use this identifier to determine
     /// which features are enabled.
-    type Spec: Debug + Copy + Send + Sync + 'static;
+    type Spec: Debug + Copy + Hash + PartialEq + Send + Sync + 'static;
```
mattsse, line 44: @klkvr `+ Into<u8>` ?

@fgimenez then we don't need to make the cache generic

klkvr: should we have `From<u8>` as well?

fgimenez: we would need it also in `revm::revm_primitives::hardfork::SpecId` and it is not available atm, should we add it there?

mattsse (approved):
lgtm

PR #89, opened 2025-05-15
[wip] feat: tracing helpers

klkvr:
## Motivation

ref https://github.com/paradigmxyz/reth/issues/16044?issue=paradigmxyz%7Creth%7C16045

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/evm/src/tracing.rs
+++ b/crates/evm/src/tracing.rs
@@ -1,0 +1,37 @@
+//! Helpers for tracing.
+
+use crate::{Evm, IntoTxEnv};
+use revm::context::result::ResultAndState;
+
+/// A helper type for tracing transactions.
+#[derive(Debug, Clone)]
+pub struct TxTracer<E: Evm> {
+    evm: E,
+    fused_inspector: E::Inspector,
+}
+
+/// Output of tracing a transaction.
+#[derive(Debug, Clone)]
+pub struct TraceOutput<H, I> {
+    /// Inner EVM output.
+    pub result: ResultAndState<H>,
+    /// Inspector state at the end of the execution.
+    pub inspector: I,
+}
+
+impl<E: Evm<Inspector: Clone>> TxTracer<E> {
+    /// Creates a new [`TxTracer`] instance.
+    pub fn new(mut evm: E) -> Self {
+        Self { fused_inspector: evm.inspector_mut().clone(), evm }
+    }
+
+    /// Executes a transaction, and returns its outcome along with the inspector state.
+    pub fn trace(
+        &mut self,
+        tx: impl IntoTxEnv<E::Tx>,
+    ) -> Result<TraceOutput<E::HaltReason, E::Inspector>, E::Error> {
+        let result = self.evm.transact(tx);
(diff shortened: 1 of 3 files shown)


crates/evm/src/tracing.rs
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    /// Executes multiple transactions, applies the closure to each transaction result, and returns
+    /// the outcomes.
+    pub fn trace_many<'a, T, O>(
+        &'a mut self,
+        txs: impl IntoIterator<Item = T, IntoIter: 'a>,
+        mut f: impl FnMut(T, ExecutionResult<E::HaltReason>, &EvmState, E::Inspector, &mut E::DB) -> O
+            + 'a,
+    ) -> impl Iterator<Item = Result<O, E::Error>> + 'a
```
mattsse, line 53: I'd like to introduce named types for these if feasible then we could expose more functionality in case we need this, and maybe stacking combinators

crates/evm/src/tracing.rs
```diff
@@ -0,0 +1,95 @@
...
+}
+
+impl<E: Evm<Inspector: Clone, DB: DatabaseCommit>> TxTracer<E> {
+    /// Creates a new [`TxTracer`] instance.
+    pub fn new(mut evm: E) -> Self {
+        Self { fused_inspector: evm.inspector_mut().clone(), evm }
+    }
+
+    fn fuse_inspector(&mut self) -> E::Inspector {
+        core::mem::replace(self.evm.inspector_mut(), self.fused_inspector.clone())
```
mattsse, line 48: this isn't quite fused,

what I'd like to support is actually fusing the `TracingInspector::fuse` because if we could reuse the instance across calls that would be a lot more efficient since we dont need to realloc a bunch of stuff.

crates/evm/src/tracing.rs
```diff
@@ -0,0 +1,95 @@
...
+        T: IntoTxEnv<E::Tx> + Clone,
+        Err: From<E::Error>,
+    {
+        txs.into_iter().map(move |tx| {
+            let result = self.evm.transact(tx.clone());
+            let inspector = self.fuse_inspector();
+            let ResultAndState { result, state } = result?;
+            let output =
+                f(TracingCtx { tx, result, state: &state, inspector, db: self.evm.db_mut() });
+            self.evm.db_mut().commit(state);
```
mattsse, line 90: this we could skip for the last tx, if we use peekable, but ideally can make this behaviour configurable, by rolling our own `Iterator` type

crates/evm/src/tracing.rs
```diff
@@ -0,0 +1,95 @@
...
+#[derive(Debug)]
+pub struct TracingCtx<'a, T, E: Evm> {
+    /// The transaction that was just executed.
+    pub tx: T,
+    /// Result of transaction execution.
+    pub result: ExecutionResult<E::HaltReason>,
+    /// State changes after transaction.
+    pub state: &'a EvmState,
+    /// Inspector state after transaction.
+    pub inspector: E::Inspector,
```
mattsse, line 27: is the problem here that we need to move ownership of the inspector into the evm in order to trace?

PR #92, opened 2025-05-19
feat: `BlockExecutor::execute_transaction_with_commit_condition`

klkvr:
## Motivation

Adds a new `execute_transaction_with_commit_condition` helper that allows to inspect the transaction outcome and commit it conditionally based on it.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/evm/src/block/mod.rs
+++ b/crates/evm/src/block/mod.rs
@@ -43,6 +43,23 @@ impl<E: BlockExecutor, T> ExecutableTx<E> for T where
 {
 }
 
+/// Marks whether transaction should be commited into block executor's state.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[must_use]
+pub enum ShouldCommit {
+    /// Transaction should be commited into block executor's state.
+    Yes,
+    /// Transaction should not be commited.
+    No,
+}
+
+impl ShouldCommit {
+    /// Returns `true` if transaction should be commited into block executor's state.
+    pub fn should_commit(self) -> bool {
+        matches!(self, Self::Yes)
+    }
+}
+
 /// A type that knows how to execute a single block.
 ///
 /// The current abstraction assumes that block execution consists of the following steps:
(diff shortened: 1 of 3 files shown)


crates/evm/src/block/mod.rs
```diff
@@ -43,6 +43,23 @@ impl<E: BlockExecutor, T> ExecutableTx<E> for T where
 {
 }
 
+/// Marks whether transaction should be commited into block executor's state.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[must_use]
+pub enum ShouldCommit {
```
mattsse, line 49: I think we should rename this because `Should` isn't definitive.

so maybe just `Commit{State|Changes}`?

crates/evm/src/block/mod.rs
```diff
@@ -81,6 +98,20 @@ pub trait BlockExecutor {
...
+        })
+    }
+
+    /// Executes a single transaction and applies execution result to internal state. Invokes the
+    /// given closure with an internal [`ExecutionResult`] produced by the EVM, and commits the
+    /// transaction to the state on [`ShouldCommit::Yes`].
+    fn execute_transaction_with_commit_condition(
+        &mut self,
+        tx: impl ExecutableTx<Self>,
+        f: impl FnOnce(&ExecutionResult<<Self::Evm as Evm>::HaltReason>) -> ShouldCommit,
```
mattsse, line 114: we could potentially make this so that we allow taking the ExecutionResult if this isn't commit::yes, but I think user could also just clone things they're interested in

mattsse (changes requested):
smol naming bikeshed

crates/evm/src/block/mod.rs
klkvr: yeah we could probably also allow overriding the result here but not sure whether it's useful

crates/op-evm/src/block/mod.rs
```diff
@@ -150,14 +154,14 @@ where
...
         // Execute transaction.
-        let result_and_state =
+        let ResultAndState { result, state } =
             self.evm.transact(tx).map_err(move |err| BlockExecutionError::evm(err, hash))?;
 
-        self.system_caller
-            .on_state(StateChangeSource::Transaction(self.receipts.len()), &result_and_state.state);
-        let ResultAndState { result, state } = result_and_state;
+        if !f(&result).should_commit() {
+            return Ok(0);
```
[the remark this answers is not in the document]
mattsse: yeah I guess we'd need to return an Either here?
at least for the conditional call?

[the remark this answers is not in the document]
klkvr: changed to `Option<u64>`

mattsse (approved):
lgtm

PR #94, opened 2025-05-20
feat: Implement `FromTxWithEncoded` and `FromRecoveredTx` from `OpTxEnvelope` for `TxEnv`

RomanHodulak:
## Motivation

While working on the `custom_node` example I intended to implement `FromTxWithEncoded<CustomTransaction> for CustomTxEnv` and the same with `FromRecoveredTx`. The approach I took was to redirect to `FromTxWithEncoded<Inner> for TxEnv`. This is, however, not implemented for `TxEnv`, but only for `OpTransaction<TxEnv>`

## Solution

Extract `impl FromTxWithEncoded<OpTxEnvelope> for TxEnv` and `FromRecoveredTx` counterpart from `impl FromTxWithEncoded<OpTxEnvelope> for OpTransaction<TxEnv>`

## PR Checklist

- [X] Added Tests
- [X] Added Documentation
- [X] Breaking changes

### DIFF
--- a/crates/evm/src/tx.rs
+++ b/crates/evm/src/tx.rs
@@ -356,13 +356,13 @@ mod op {
     use op_revm::{transaction::deposit::DepositTransactionParts, OpTransaction};
     use revm::context::TxEnv;
 
-    impl FromTxWithEncoded<OpTxEnvelope> for OpTransaction<TxEnv> {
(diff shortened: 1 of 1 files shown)


mattsse (approved):
I believe this makes sense and isn't problematic because this wouldn't be usable with op-evm anyway 

pending @klkvr

klkvr (approved):
makes sense to me

PR #102, opened 2025-06-16
feat: move map_precompiles to trait to allow precompile customization

debjit-bw:
## Motivation

Replacing the hardcoded PrecompilesMap with the impl representations allows us to drop the hardcoded PrecompilesMap in reth's ConfigureEvm. This helps to extend the precompile's usecase to allow custom processing. Currently the PrecompileProvider, despite implementation, isn't used for custom precompiles (which was allowed earlier).

Related discussion: https://github.com/paradigmxyz/reth/discussions/16825

## Solution

Created a new trait called `PrecompileCfg` which includes the functions used by reth to set the precompiles.

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [x] Breaking changes

### DIFF
--- a/crates/evm/src/eth/mod.rs
+++ b/crates/evm/src/eth/mod.rs
@@ -228,13 +228,13 @@ where
 pub struct EthEvmFactory;
 
 impl EvmFactory for EthEvmFactory {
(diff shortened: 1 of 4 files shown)


mattsse:
yeah, we decided to stick with moving more things to `PrecompileInput` e.g. access to journal and blockenv

PR #109, opened 2025-06-19
feat!: provide more context to `Precompile::call`

klkvr:
## Motivation

ref https://github.com/paradigmxyz/reth/discussions/16825#discussioncomment-13490779

## Solution

Changes `Precompile::call` input to be a single struct with context for precompile call including caller address and call value.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mattsse:
we can keep #102 open, in case we also need to do this

PR #126, opened 2025-07-03
feat: add is_pure method to Precompile trait

### LINKED ISSUE #125: [Feature] Add Precompile::is_stateless

we want to support precompiles that need to operate on the current state.

usually precompile output is deterministic based on the input, and therefor cacheable.

but we need a way to opt-out of this by marking it as non determinsitc

## TODO
* add Precompile::is_stateless -> bool
* false for all `fn`

* for closures we can also make it false imo

### Additional context

_No response_

mattsse:
## Summary

Implements the `is_pure` method for the `Precompile` trait as described in #125. This allows precompiles to declare whether their output is deterministic based solely on input, enabling potential caching optimizations.

## Changes

- Added `is_pure` method to the `Precompile` trait with a default implementation returning `false`
- Updated trait implementations to properly delegate `is_pure` calls:
  - `DynPrecompile` and `&DynPrecompile` now delegate to their inner precompile
  - `Either<A, B>` delegates to the appropriate variant
- Added comprehensive documentation explaining the purpose and usage of `is_pure`
- Added tests to verify the implementation works correctly for all trait implementations

The conservative default of `false` ensures existing precompiles that depend on state continue to work without modification. Precompiles can opt into being marked as pure by overriding the method to return `true`.

Fixes #125

### DIFF
--- a/crates/evm/src/precompiles.rs
+++ b/crates/evm/src/precompiles.rs
@@ -342,6 +342,20 @@ impl DynPrecompile {
     {
         Self(Arc::new(f))
     }
+
+    /// Creates a new [`DynPrecompiles`] with the given closure and [`Precompile::is_pure`]
+    /// returning `false`.
+    pub fn new_stateful<F>(f: F) -> Self
+    where
+        F: Fn(PrecompileInput<'_>) -> PrecompileResult + Send + Sync + 'static,
+    {
+        Self(Arc::new(StatefulPrecompile(f)))
+    }
+
+    /// Flips [`Precompile::is_pure`] to `false`.
+    pub fn stateful(self) -> Self {
+        Self(Arc::new(StatefulPrecompile(self.0)))
+    }
 }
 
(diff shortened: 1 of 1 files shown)


crates/evm/src/precompiles.rs
```diff
@@ -387,6 +387,37 @@ pub struct PrecompileInput<'a> {
...
+    ///     }
+    ///
+    ///     fn is_pure(&self) -> bool {
+    ///         true // This precompile always produces the same output for the same input
+    ///     }
+    /// }
+    /// ```
+    fn is_pure(&self) -> bool {
+        false
+    }
```
klkvr, line 420: should we default to true here?

klkvr (commented):
I'm wondering if a builder-style helper for flipping this on `DynPrecompile` could be useful? like `DynPrecompile::from(|input| ...).stateful()`

or just a constructor for `DynPrecompile` that would set this

PR #127, opened 2025-07-06
feat: add helper EvmFactory type that modifies the PrecompilesMap

### LINKED ISSUE #119: [Feature] Add helper EvmFactory type that modifies the PrecompilesMap

we have

https://github.com/alloy-rs/evm/blob/b83a678656e6ebdebf4f96be69d1ea6294e67dc6/crates/evm/src/eth/mod.rs#L239-L239

it would be useful to have some generic type that takes a default EvmFactory and takes care of the PrecompileMap init based in the specid

this can be a generic EvmFactoryWith<Evm, PrecompilesInit>

where PrecompilesInit takes care of modifiying the configured precompilesmap

https://github.com/alloy-rs/evm/blob/b83a678656e6ebdebf4f96be69d1ea6294e67dc6/crates/evm/src/eth/mod.rs#L256-L258

something like

```
trait PrecompileModifies<SpecId> {

  fn modify(&self, precompiles: &mut PrecompilesMap, specid: SpecId)

}
```

### Additional context

_No response_

stevencartavia:
## Motivation
closes #119

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/evm/src/eth/mod.rs
+++ b/crates/evm/src/eth/mod.rs
@@ -281,56 +281,6 @@ impl EvmFactory for EthEvmFactory {
     }
 }
 
-/// Trait for modifying precompiles based on spec ID.
-pub trait PrecompileModifies<SpecId> {
-    /// Modifies the precompiles based on the spec ID.
-    fn modify(&self, precompiles: &mut PrecompilesMap, spec_id: SpecId);
-}
-
-/// A wrapper around an [`EvmFactory`] that modifies the precompiles based on the spec ID.
-#[derive(Debug, Clone, Copy)]
-pub struct EvmFactoryWith<Evm, PrecompilesInit> {
-    factory: Evm,
-    precompile_init: PrecompilesInit,
-}
-
-impl<F, P> EvmFactory for EvmFactoryWith<F, P>
-where
-    F: EvmFactory<Precompiles = PrecompilesMap, Spec = SpecId>,
-    P: PrecompileModifies<SpecId>,
-{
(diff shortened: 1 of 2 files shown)


crates/evm/src/eth/mod.rs
```diff
@@ -281,6 +281,56 @@ impl EvmFactory for EthEvmFactory {
     }
 }
 
+/// Trait for modifying precompiles based on spec ID.
+pub trait PrecompileModifies<SpecId> {
```
klkvr, line 285: let's add a blanket impl for `Fn(&mut PrecompilesMap, SpecId)`

klkvr (changes requested):
let's move this to `crates/evm/src/evm.rs` and add a helper 
```rust
fn with_modified_precompiles<F>(self, f: F) -> EvmFactoryWith<Self, F> 
where 
    F: Fn(&mut PrecompilesMap, Self::Spec)
``` 

to `EvmFactoryExt` here: 
https://github.com/alloy-rs/evm/blob/fb075bf431e569d81eb2ed24ba604b4be2d76a1f/crates/evm/src/evm.rs#L240

klkvr:
I'm also wondering if we shuld have a more general `trait EvmOverrides` that would take `&mut F::Evm`? that way `PrecompileOverrides` can be just a subset of it while allowing for e.g `EvmFactoryExt::with_evm_overrides`

wdyt @mattsse

mattsse:
@klkvr should this then be called `trait EvmExt`?

mattsse (commented):
decided against doing this

PR #131, opened 2025-07-11
feat: make fusing optional in `TxTracer`

### LINKED ISSUE #130: [Feature] Make cloning in TracerIter optional

currently we fuse (clone) the inspector each time

https://github.com/alloy-rs/evm/blob/9d831c815b41dd610ef28b6248556b3a61b3b750/crates/evm/src/tracing.rs#L129-L129

we could avoid this by giving the consume an option to fuse the inspector manually,
this could look similar to binaryheap::PeekMut

where we instead of doing

https://github.com/alloy-rs/evm/blob/9d831c815b41dd610ef28b6248556b3a61b3b750/crates/evm/src/tracing.rs#L129-L129
then
https://github.com/alloy-rs/evm/blob/9d831c815b41dd610ef28b6248556b3a61b3b750/crates/evm/src/tracing.rs#L137-L137

we do something like

```rust
struct TraceMut {
  &mut TxTracer
}

fn take_inspector() {
  self.fuse_inspector()
}

fn inspector_mut(&mut) {
  // user can manually fuse
}
```

this needs some extra care so that we can both access

https://github.com/alloy-rs/evm/blob/9d831c815b41dd610ef28b6248556b3a61b3b750/crates/evm/src/tracing.rs#L138-L138

and the inspector at the same time, and ideally some safety that the user must op-out of cloning

### Additional context

_No response_

klkvr:
## Motivation

Closes #130 

## Solution

Adds `no_fuse` method to `TracerIter` that allows to disable fusing of inspector and rely on user to do it manually.

Changes `Evm` trait to expose just `components` and `components_mut` allowing to borrow all components at the same time instead of requiring all implementations to define separate getters for db/inspector/precompiles

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/evm/src/tracing.rs
+++ b/crates/evm/src/tracing.rs
@@ -25,9 +25,18 @@ pub struct TracingCtx<'a, T, E: Evm> {
     /// State changes after transaction.
     pub state: &'a EvmState,
     /// Inspector state after transaction.
-    pub inspector: E::Inspector,
+    pub inspector: &'a mut E::Inspector,
     /// Database used when executing the transaction, _before_ committing the state changes.
     pub db: &'a mut E::DB,
+    /// Fused inspector.
+    fused_inspector: &'a E::Inspector,
+}
+
+impl<'a, T, E: Evm<Inspector: Clone>> TracingCtx<'a, T, E> {
+    /// Fuses the inspector and returns the current inspector state.
+    pub fn take_inspector(&mut self) -> E::Inspector {
+        core::mem::replace(self.inspector, self.fused_inspector.clone())
+    }
 }
 
 /// Output of tracing a transaction.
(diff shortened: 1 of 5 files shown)


crates/op-evm/src/lib.rs
```diff
@@ -202,20 +198,20 @@ where
...
+        (
+            &self.inner.0.ctx.journaled_state.database,
+            &self.inner.0.inspector,
+            &self.inner.0.precompiles,
+        )
     }
 
-    fn inspector_mut(&mut self) -> &mut Self::Inspector {
-        &mut self.inner.0.inspector
+    fn components_mut(&mut self) -> (&mut Self::DB, &mut Self::Inspector, &mut Self::Precompiles) {
```
mattsse, line 209: I think this makes sense, because we expect that an evm wraps all of these

crates/evm/src/tracing.rs
```diff
@@ -25,9 +25,18 @@ pub struct TracingCtx<'a, T, E: Evm> {
     /// State changes after transaction.
     pub state: &'a EvmState,
     /// Inspector state after transaction.
-    pub inspector: E::Inspector,
+    pub inspector: &'a mut E::Inspector,
     /// Database used when executing the transaction, _before_ committing the state changes.
     pub db: &'a mut E::DB,
+    /// Fused inspector.
+    fused_inspector: &'a E::Inspector,
```
mattsse, line 32: there's still a mandatory clone here, which wouldn't work for 

https://github.com/paradigmxyz/revm-inspectors/blob/eb6ed5965050e8723ccf1ec02b0b53f7f0005af5/src/tracing/js/mod.rs#L56-L57

for example.

but what we can do next is replace this with something that returns a new inspector, like `Setup = Fn() -> E::Inspector` which by default could be `|| ins.clone()`

mattsse (approved):
nice, 

smol suggestion for followup

PR #137, opened 2025-07-16
feat: enhance precompile lookup documentation and functionality

mattsse:
Improves the precompile lookup functionality with better documentation and proper implementation.

## Changes

- Enhanced documentation for `set_precompile_lookup` and `with_precompile_lookup` methods with detailed explanations and examples
- Updated `get()` method to properly check the lookup function using nested `Either` types
- Added `PrecompileLookup` type alias for improved code readability
- Added comprehensive test coverage for dynamic precompile lookup functionality

## Test Plan

Added new test `test_precompile_lookup` that verifies:
- Static precompiles continue to work normally
- Dynamic lookup function is called for addresses not in the static map
- Lookup function correctly returns precompiles for matching addresses
- Non-matching addresses return `None`

All existing tests continue to pass.

### DIFF
--- a/crates/evm/src/precompiles.rs
+++ b/crates/evm/src/precompiles.rs
@@ -16,7 +16,6 @@ use revm::{
     Context, Journal,
 };
 
-
 /// A mapping of precompile contracts that can be either static (builtin) or dynamic.
 ///
 /// This is an optimization that allows us to keep using the static precompiles


crates/evm/src/precompiles.rs
```diff
@@ -16,6 +16,8 @@ use revm::{
     Context, Journal,
 };
 
+type PrecompileLookup = Arc<dyn Fn(&Address) -> Option<DynPrecompile> + Send + Sync>;
```
klkvr, line 19: should this be return &DynPrecompile?

mattsse: maybe yes, but this would then require that that this lookup always owns the precompile
and since this is just an Arc clone worst case this is likely fine, but &Dyn would be nicer indeed, so we can change this to enforce some design decisions for this

klkvr: yeah I think ideally you'd just keep those as `&'static DynPrecompile`

mattsse: actually, this is introducing some lifetime issues...

klkvr: I guess we should either have a trait for this or lock into 'static references

klkvr (approved):
nice

PR #138, opened 2025-07-17
feat: Introduced Initializers to trace inspector

### LINKED ISSUE #132: [Feature] Use initializer for Inspector reset

we currently rely on clone and do a mandatory clone to store the initial state:

https://github.com/alloy-rs/evm/blob/1304499b4cde025b1aecae5579403d6391831e7f/crates/evm/src/tracing.rs#L43-L46

this does not work for non clone inspectors:

https://github.com/paradigmxyz/revm-inspectors/blob/eb6ed5965050e8723ccf1ec02b0b53f7f0005af5/src/tracing/js/mod.rs#L56-L57

instead we can introduce an `Init` generic that is

```
FnMut() ->  E::Inspector
```

and by default 

```
let fused = ins.clone();
let init = move || ins.clone();
```

### Additional context

_No response_

developeruche:
Close #132

## Solution
Introduced an `InspectorInitializer` which can init both cloneable and non-cloneable inspectors

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/evm/src/tracing.rs
+++ b/crates/evm/src/tracing.rs
@@ -1,3 +1,191 @@
+// //! Helpers for tracing.
+
+// use crate::{Evm, IntoTxEnv};
+// use core::{fmt::Debug, iter::Peekable};
+// use revm::{
+//     context::result::{ExecutionResult, ResultAndState},
+//     state::EvmState,
+//     DatabaseCommit,
+// };
+
+// /// A helper type for tracing transactions.
+// #[derive(Debug, Clone)]
+// pub struct TxTracer<E: Evm> {
+//     evm: E,
+//     fused_inspector: E::Inspector,
+// }
+
+// /// Container type for context exposed in [`TxTracer`].
+// #[derive(Debug)]
+// pub struct TracingCtx<'a, T, E: Evm> {
+//     /// The transaction that was just executed.
+//     pub tx: T,
+//     /// Result of transaction execution.
+//     pub result: ExecutionResult<E::HaltReason>,
+//     /// State changes after transaction.
+//     pub state: &'a EvmState,
+//     /// Inspector state after transaction.
+//     pub inspector: &'a mut E::Inspector,
+//     /// Database used when executing the transaction, _before_ committing the state changes.
+//     pub db: &'a mut E::DB,
+//     /// Fused inspector.
+//     fused_inspector: &'a E::Inspector,
+//     /// Whether the inspector was fused.
+//     was_fused: &'a mut bool,
+// }
+
+// impl<'a, T, E: Evm<Inspector: Clone>> TracingCtx<'a, T, E> {
+//     /// Fuses the inspector and returns the current inspector state.
+//     pub fn take_inspector(&mut self) -> E::Inspector {
+//         *self.was_fused = true;
+//         core::mem::replace(self.inspector, self.fused_inspector.clone())
+//     }
+// }
+
+// impl<E: Evm<Inspector: Clone, DB: DatabaseCommit>> TxTracer<E> {
+//     /// Creates a new [`TxTracer`] instance.
+//     pub fn new(mut evm: E) -> Self {
+//         Self { fused_inspector: evm.inspector_mut().clone(), evm }
+//     }
+
+//     fn fuse_inspector(&mut self) -> E::Inspector {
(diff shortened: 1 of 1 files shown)


mattsse:
> Initialize should return a result rather than the plain inspector?

yep, for example if initializer is `|| JsInpesctor::new()`

crates/evm/src/tracing.rs
```diff
@@ -1,23 +1,78 @@
...
-use core::{fmt::Debug, iter::Peekable};
+use core::{fmt::Debug, iter::Peekable, marker};
 use revm::{
     context::result::{ExecutionResult, ResultAndState},
     state::EvmState,
     DatabaseCommit,
 };
 
+/// Trait for inspector initialization.
+pub trait InspectorInitializer<T, E> {
```
klkvr, line 12: `E` should be a `type Error` associated type

crates/evm/src/tracing.rs
```diff
@@ -1,23 +1,78 @@
...
+    /// Create a new inspector instance.
+    fn initialize(&mut self) -> Result<T, E>;
+}
+
+/// A simple initializer that clones the inspector.
+#[derive(Debug, Clone)]
+pub struct CloneInitializer<T: Clone, E>(T, marker::PhantomData<E>);
```
klkvr, line 19: this would then allow to remove the phantomdata from here and have simply `type Error = Infallible` instead

crates/evm/src/tracing.rs
```diff
@@ -1,23 +1,78 @@
...
+    pub fn new(f: F) -> Self {
+        Self(f)
+    }
+}
+
+impl<T, E, F: FnMut() -> Result<T, E>> InspectorInitializer<T, E> for FnInitializer<T, E, F> {
+    fn initialize(&mut self) -> Result<T, E> {
+        (self.0)()
+    }
+}
```
klkvr, line 49: could we implement this for `F: FnMut() -> Result<T, E>` directly?

crates/evm/src/tracing.rs
```diff
@@ -1,23 +1,78 @@
...
+
 /// A helper type for tracing transactions.
 #[derive(Debug, Clone)]
-pub struct TxTracer<E: Evm> {
+pub struct TxTracer<
+    E: Evm,
+    Init: InspectorInitializer<E::Inspector, E::Error> = CloneInitializer<
+        <E as Evm>::Inspector,
+        <E as Evm>::Error,
```
klkvr, line 57:
```suggestion
        E::Inspector,
        E::Error,
```

crates/evm/src/tracing.rs
```diff
@@ -1,23 +1,78 @@
...
 /// Container type for context exposed in [`TxTracer`].
 #[derive(Debug)]
-pub struct TracingCtx<'a, T, E: Evm> {
+pub struct TracingCtx<
+    'a,
+    T,
+    E: Evm,
+    Init: InspectorInitializer<E::Inspector, E::Error> = CloneInitializer<
+        <E as Evm>::Inspector,
+        <E as Evm>::Error,
```
klkvr, line 73:
```suggestion
        E::Inspector,
        E::Error,
```

crates/evm/src/tracing.rs
```diff
@@ -32,33 +87,48 @@ pub struct TracingCtx<'a, T, E: Evm> {
...
-    /// Fuses the inspector and returns the current inspector state.
-    pub fn take_inspector(&mut self) -> E::Inspector {
+impl<'a, T, E: Evm, Init: InspectorInitializer<E::Inspector, E::Error>> TracingCtx<'a, T, E, Init> {
+    /// Takes the current inspector and replaces it with a fresh one.
+    ///
+    /// This is useful when you want to keep the current inspector state but continue
+    /// tracing with a new one.
+    pub fn take_inspector(&mut self) -> Result<E::Inspector, E::Error>
+    where
+        E::Inspector: Clone,
```
klkvr, line 101: can we avoid this bound?

klkvr (changes requested):

crates/evm/src/tracing.rs
developeruche: yeah

PR #139, opened 2025-07-18
perf: allocate capacity for balance changes

mattsse:
we know how many entries we need

### DIFF
--- a/crates/evm/src/block/state_changes.rs
+++ b/crates/evm/src/block/state_changes.rs
@@ -26,7 +26,10 @@ pub fn post_block_balance_increments<H>(
 where
     H: BlockHeader,
 {
-    let mut balance_increments = HashMap::default();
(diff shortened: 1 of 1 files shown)


crates/evm/src/block/state_changes.rs
```diff
@@ -26,7 +26,10 @@ pub fn post_block_balance_increments<H>(
 where
     H: BlockHeader,
 {
-    let mut balance_increments = HashMap::default();
+    let mut balance_increments = HashMap::with_capacity_and_hasher(
+        withdrawals.map_or(ommers.len(), |w| w.len()),
```
DaniPopes, line 30: isn't this withdrawals + ommers?

DaniPopes (approved):

crates/evm/src/block/state_changes.rs
DaniPopes: right

PR #145, opened 2025-07-29
refactor: simplify `BlockExecutorFactory` lifetimes

klkvr:
## Motivation

Adds `StateDB` trait that erases lifetime of `&'a mut State<DB>` and simplifies bounds on `create_executor`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/evm/src/block/mod.rs
+++ b/crates/evm/src/block/mod.rs
@@ -1,12 +1,15 @@
 //! Block execution abstraction.
 
+use core::fmt::Debug;
+
 use crate::{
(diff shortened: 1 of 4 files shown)


mattsse (approved):
makes sense,

this does look cleaner and more or less still locks in the State type  although now we could also get rid of that in case this is required

mattsse:
pedantic clippy

klkvr:
solved differently on reth side

PR #146, opened 2025-07-29
feat: add system call inspection support for EVM tracing

rezbera:
This PR adds system call inspection capabilities to alloy-evm by implementing the `InspectSystemCallEvm` trait and updating the `SystemCallEvm` implementation. The changes enable proper tracing and debugging of system calls during EVM execution, which is essential for comprehensive transaction and block tracing functionality. The implementation uses revm's inspection framework to capture system call execution details while maintaining compatibility with existing EVM execution flows.

builds on https://github.com/bluealloy/revm/pull/2808


Cargo.toml
```diff
@@ -63,6 +63,18 @@ serde = { version = "1", default-features = false, features = ["derive"] }
...
+revm-inspector = { git = "https://github.com/bluealloy/revm", rev = "9008e93" }
+revm-handler = { git = "https://github.com/bluealloy/revm", rev = "9008e93" }
+revm-context = { git = "https://github.com/bluealloy/revm", rev = "9008e93" }
+revm-context-interface = { git = "https://github.com/bluealloy/revm", rev = "9008e93" }
+revm-database = { git = "https://github.com/bluealloy/revm", rev = "9008e93" }
+revm-database-interface = { git = "https://github.com/bluealloy/revm", rev = "9008e93" }
+revm-state = { git = "https://github.com/bluealloy/revm", rev = "9008e93" }
+revm-bytecode = { git = "https://github.com/bluealloy/revm", rev = "9008e93" }
+revm-precompile = { git = "https://github.com/bluealloy/revm", rev = "9008e93" }
+op-revm = { git = "https://github.com/bluealloy/revm", rev = "9008e93" }
```
rezbera, line 80: todo use actual release for revm when its ready

rezbera:
> if self.inspect {
>             self.inner.inspect_system_call_with_caller(caller, contract, data)
>         } else {
>             self.inner.system_call_with_caller(caller, contract, data)
>         }

jank workaround for berachain. will need to figure out a real solution

mattsse (changes requested):
hmm, the evm itself, doesnt really have a notion of inspecting transaction,
due to how this is implemented it is assumed that this is controlled by the inspector toggle.

@rezbera I believe you're using the regular eth evm right?

could we perhaps make this an EthEvm native function first before we embedd this in the trait?

rezbera:
closing as it makes more sense to just have our own Evm

PR #148, opened 2025-07-30
feat: `ToTxEnv`

klkvr:
## Motivation

The way we've designed and use most of the APIs is by passing transactions directly to EVM/block executors, thus using `IntoTxEnv` does not make perfect sense as it consumes `self` while we're mostly doing `&self -> TxEnv` conversions now.

## Solution

This PR introduces `BuildTxEnv` which takes `&self` and is automatically implemented for references. This allows to simplify some of the `TxEnv` implementations and drop `Copy` supertrait from `ExecutableTx`, allowing to pass owned transactions to block executors instead of just references.

The only `IntoTxEnv` implementations are now for `BuildTxEnv` and for `Self`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/evm/src/block/mod.rs
+++ b/crates/evm/src/block/mod.rs
@@ -1,7 +1,7 @@
 //! Block execution abstraction.
 
 use crate::{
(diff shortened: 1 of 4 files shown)


mattsse (approved):
lgtm

PR #149, opened 2025-07-30
feat: add evmext trait

mattsse:
Adds some common evm transact features.

I think we would benefit from a standalone trait that provides the txhash

the alternative rn would be via Encodable2718


crates/evm/src/evm.rs
```diff
@@ -191,6 +191,69 @@ pub trait Evm {
...
+            index += 1;
+        }
+        Ok(index)
+    }
+
+    /// Replays all the previous transactions and returns the [`ResultAndState`] of the target
+    /// transaction.
+    ///
+    /// Returns `None` if the target transaction was not found.
+    fn replay_transaction<I, T>(
```
klkvr, line 232: would it make sesnse for this to disable inspection while executing transactions before the target one? assuming that what we need here is to only trace it

klkvr: I don't think there's a way for caller to prevent inspector from being called for all transactions before it

rn it's either not called at all, or called for all of the executed txs

klkvr (approved):

PR #153, opened 2025-08-15
feat: add map_pure_precompiles to respect pure precompiles for caching

yongkangc:
Implements `map_pure_precompiles` to filter out non-pure precompiles that may have side effects when caching. Adds helper method `map_precompiles_filtered` for conditional precompile transformations.

Closes paradigmxyz/reth#17870


mattsse (approved):
lgtm!

PR #155, opened 2025-08-15
feat(eth): introduce EthEvmBuilder for unified EthEvm instance creation

### LINKED ISSUE #154: [Feature] Add EthEvmbuilder helper type

we have both: https://github.com/alloy-rs/evm/blob/4a1dbbbeefc8832c8b22a68ff7c804f3d5647358/crates/evm/src/eth/mod.rs#L185-L185

and 

https://github.com/alloy-rs/evm/blob/4a1dbbbeefc8832c8b22a68ff7c804f3d5647358/crates/evm/src/eth/mod.rs#L200-L200

it would be nice to have a helper builder type that wraps

https://github.com/alloy-rs/evm/blob/4a1dbbbeefc8832c8b22a68ff7c804f3d5647358/crates/evm/src/eth/mod.rs#L208-L215

and allows us to unify this logic, by basically abstracting over `Context::mainnet()`

as input we need mandatory for building the evm

* blockenv
* cfg
* db

optional setters for
* inspector
* inspect
* precompilesmap

## TODO
* add this helper builder type

### Additional context

_No response_

mablr:
## Motivation

Close #154 

## Solution

Introducing `EthEvmBuilder` struct helper abstracting over `Context::mainnet()`.

Mandatory inputs:
 - `DB`
 - `EvmEnv`
   - `BlockEnv`
   - `CfgEnv`

Optional setters:
 - `inspector`
 - `inspect` (whether to invoke the inspector on transact)
 - `precompiles` (custom `PrecompilesMap` override)
 
## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/evm/src/eth/mod.rs
+++ b/crates/evm/src/eth/mod.rs
@@ -41,7 +41,7 @@ pub struct EthEvmBuilder<DB: Database, I = NoOpInspector> {
 
 impl<DB: Database> EthEvmBuilder<DB, NoOpInspector> {
     /// Creates a builder from the provided `EvmEnv` and database.
-    pub fn from_env(db: DB, env: EvmEnv) -> Self {
+    pub fn new(db: DB, env: EvmEnv) -> Self {
         Self {
             db,
             block_env: env.block_env,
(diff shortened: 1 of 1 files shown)


crates/evm/src/eth/mod.rs
```diff
@@ -28,6 +28,80 @@ pub mod spec;
...
+            inspector: NoOpInspector {},
+            inspect: false,
+            precompiles: None,
+        }
+    }
+}
+
+impl<DB: Database, I> EthEvmBuilder<DB, I> {
+    /// Sets a custom inspector and changes the inspector type of the builder.
+    pub fn inspector<J>(self, inspector: J) -> EthEvmBuilder<DB, J> {
```
mattsse, line 58: here we should add a note that this only installs the inspector but does not enabled it

we can perhaps add `fn activate_inspector<J>(self, inspector: J) -> EthEvmBuilder<DB, J> {`

that does both

PR #160, opened 2025-08-25
feat: add State generic to Evm trait

alessandromazza98:
### Problem

- as of now the `Evm` trait hard codes the `ResultAndState<Self::HaltReason>` without providing the possibility to use a custom state, different than the evm state.
- The `ResultAndState` though has the possibility to use a generic `S` (it defaults to `EvmState`).
- This way it's not possible to use a custom state in the evm.

### Solution

- add a new associated type `State: Into<EvmState>` and use it in the `ResultAndState` type: `ResultAndState<Self::HaltReason, Self::State>`.
- we still need the `State` associated type to be convertible into a `EvmState` now, but at least we let implementers of the `Evm` trait to use a custom state that wraps the `EvmState` with something more.

What do you think? @mattsse

### DIFF
--- a/crates/evm/src/block/system_calls/eip2935.rs
+++ b/crates/evm/src/block/system_calls/eip2935.rs
@@ -8,7 +8,7 @@ use alloc::string::ToString;
 use alloy_eips::eip2935::HISTORY_STORAGE_ADDRESS;
 use alloy_hardforks::EthereumHardforks;
 use alloy_primitives::B256;
-use revm::context_interface::result::ResultAndState;
+use revm::{context_interface::result::ResultAndState, state::EvmState};
 
 /// Applies the pre-block call to the [EIP-2935] blockhashes contract, using the given block,
 /// chain specification, and EVM.
@@ -23,11 +23,11 @@ use revm::context_interface::result::ResultAndState;
 ///
 /// [EIP-2935]: https://eips.ethereum.org/EIPS/eip-2935
 #[inline]
-pub(crate) fn transact_blockhashes_contract_call<Halt>(
+pub(crate) fn transact_blockhashes_contract_call<Halt, State: Into<EvmState>>(
     spec: impl EthereumHardforks,
     parent_block_hash: B256,
-    evm: &mut impl Evm<HaltReason = Halt>,
-) -> Result<Option<ResultAndState<Halt>>, BlockExecutionError> {
+    evm: &mut impl Evm<HaltReason = Halt, State = State>,
+) -> Result<Option<ResultAndState<Halt, State>>, BlockExecutionError> {
     if !spec.is_prague_active_at_timestamp(evm.block().timestamp.saturating_to()) {
         return Ok(None);
     }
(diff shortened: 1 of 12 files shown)


alessandromazza98:
> what's an example for this?
> 
> with still we still only can commit the evmstate so this the expectation that this is like `MyState {EvmState, extrastuff}`?

correct, basically it lets implementers to use a custom state that wraps the evm state. Your example perfectly fits.

This is important because then in the `BlockExecutor` trait, when we call the `evm.transact(tx)` it returns this custom state. And then we can both commit the evm and use the extra stuff for other purposes.

alessandromazza98:
> Do you have any specific usecase that this will enable? I could see how generic state would allow returning more context from EVM, however currently we're still bottlenecked by the database APIs I believe? i.e you can't commit anything besides `EvmState` to `DatabaseCommit` impl meaning that any custom state transitions can't be applied.
> 
> The way I've thought about state abstraction since the begging was that the key abstraction point will be an `Account` because state is basically an `Address -> Account` mapping, and adding other dimensions to it might not be very useful

sure I can bring my example here.

I use a custom evm that has a custom state. This custom state is something like:

```rust
struct MyEvmState {
    evm_state: EvmState,
    extra_state: ExtraState
}

// this struct contains some data that I need but you can imagine it to
// just be something like this.
struct ExtraState {
    counter: u64,
}
```

Then in the `ExecuteEvm` trait implementation that is used when calling `evm.transact(tx)` I constrain the associated `State` type to be this `MyEvmState`. Note here that revm already lets you use a custom `State` type here...

Then, when I want to use my custom evm in the executor of the node, I need to also impl the `BlockExecutor` trait so that it can not only execute the tx on the evm, but create the receipts, commit changes etc... And inside this trait impl I'd like to use the same custom state I used in the `ExecuteEvm` trait.

But right now, without this PR, I cannot do it because I'm constrained to use an evm that implements this `Evm` trait and this trait doesn't let me customize the state that is returned because it's hard coded to be the `EvmState`.

But ideally I'd like this PR to be merged so that in the `BlockExecutor` trait impl I can then do this:

```rust
impl BlockExecutor for MyBlockExecutor {
    fn execute_transaction_with_commit_condition(
        &mut self,
        tx: impl ExecutableTx<Self>,
        f: impl FnOnce(&ExecutionResult<<Self::Evm as Evm>::HaltReason>) -> CommitChanges,
    ) -> Result<Option<u64>, BlockExecutionError> {
        // ... some code here for gas checks and so on

        // execute the tx
        let ResultAndState {
                result,
                state: my_custom_state,
            } = self.evm.transact(tx)?        

         // get evm state
        let evm_state = my_custom_state.evm_state;
        // commit them to the db
        self.evm.db_mut().commit(evm_state);
        // get my extra state
        let extra_state = my_custom_state.extra_state;
        // and then do whatever I want here with my extra state.
        // For example I can increment a counter in MyBlockExecutor that is used somewhere in the code
        self.counter += extra_state.counter;
        
        // ... rest of the code
    }
}
```

klkvr:
yeah but what would this abstraction use enable exactly? is it required to re-implement some existing EVM? I don't think this pattern can be integrated into e.g reth right now

mattsse:
is this an extra byproduct that you need to include in the block?

PR #163, opened 2025-08-26
feat: decompose execute_transaction_with_commit_condition in BlockExecutor

### LINKED ISSUE #162: [Feature] Decompose execute_transaction_with_commit_condition in Blockexecutor

we want a way to commit/append executed transactions directly, e.g. if the result was obtained seperately, for this we can decompose execute_transaction_with_commit_condition into execute_transaction -> Output and commit_transaction(output: Output) , where output is either a new AT or just ResultAndState for now

### Additional context

we have db_mut().commit on evm but we dont have something like this on the blockexecutor
we should probably add something like this
https://github.com/alloy-rs/evm/blob/a2c1441a505f33979f14eac183a606646d893ce0/crates/evm/src/eth/block.rs#L157-L162
but as like commit_transaction that only does
https://github.com/alloy-rs/evm/blob/a2c1441a505f33979f14eac183a606646d893ce0/crates/evm/src/eth/block.rs#L184-L196

yongkangc:
## Summary

Implements the decomposition of `execute_transaction_with_commit_condition` into two separate methods as requested in #162:
- `execute_transaction_without_commit` - Executes transaction without committing state changes
- `commit_transaction` - Commits a previously executed transaction's output

## Changes

- Added `TransactionOutput` type alias for `ResultAndState` to encapsulate execution output
- Added two new methods to `BlockExecutor` trait with documentation
- Implemented methods in `EthBlockExecutor` and `OpBlockExecutor`
- Refactored `execute_transaction_with_commit_condition` as a default trait method using the new decomposed methods
- Updated all transaction execution methods to require `Copy` bound for consistency

## Benefits

1. **Separation of Concerns**: Execution and commit are now separate operations
2. **Flexibility**: Users can execute transactions without committing (useful for simulations)
3. **Consistency**: Follows the same pattern as the EVM's `db_mut().commit()` method
4. **Backward Compatibility**: Existing method continues to work using the new methods internally
5. **Testability**: Easier to test execution and commit logic separately

## Test Plan

- [x] Code compiles without errors
- [x] All clippy lints pass
- [x] Code formatted with cargo fmt
- [ ] Existing tests pass
- [ ] Add new tests for the decomposed methods

Closes #162


mattsse (commented):
hmm, not sure if this is the direction we want to take this and lock in & (Copy) args here.

I wonder if we should leave the execute_transaction_with_commit_condition as is and only add the commit_execute as new fn instead

crates/evm/src/block/mod.rs
```diff
@@ -193,9 +200,52 @@ pub trait BlockExecutor {
...
+    ///
+    /// Returns the gas used by the transaction.
+    ///
+    /// # Parameters
+    /// - `output`: The transaction output containing execution result and state changes
+    /// - `tx`: The original transaction (needed for receipt generation)
+    fn commit_transaction(
+        &mut self,
+        output: TransactionOutput<<Self::Evm as Evm>::HaltReason>,
+        tx: impl ExecutableTx<Self> + Copy,
```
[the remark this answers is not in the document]
yongkangc: this makes sense, thanks matt for the suggestion

crates/evm/src/block/mod.rs
```diff
@@ -197,6 +197,44 @@ pub trait BlockExecutor {
...
+    ///
+    /// # Parameters
+    /// - `output`: The transaction output containing execution result and state changes
+    /// - `tx`: The original transaction (needed for receipt generation)
+    fn commit_transaction(
+        &mut self,
+        output: revm::context_interface::result::ResultAndState<<Self::Evm as Evm>::HaltReason>,
+        tx: &impl ExecutableTx<Self>,
+    ) -> Result<u64, BlockExecutionError>;
+
```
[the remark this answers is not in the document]
yongkangc: thanks for the suggestion, that does sound good! thanks @mattsse and @klkvr

crates/op-evm/src/block/mod.rs
```diff
@@ -214,6 +217,99 @@ where
...
+            .into());
+        }
+
+        let hash = tx.tx().trie_hash();
+
+        // Execute transaction and return the result
+        self.evm.transact(tx).map_err(move |err| BlockExecutionError::evm(err, hash))
+    }
+
+    fn commit_transaction(
```
yongkangc, line 243: @mattsse @klkvr is this an issue for op where we need to have the commit be stateful?

what i am thinking of right now is to override the main solution.

we can either pass in the cached nonce (which would break trait) or we can have it in block executor (which is ugly and makes it stateful to a txn)

what do you guys think? is there a better alternative?

crates/op-evm/src/block/mod.rs
```diff
@@ -214,6 +217,99 @@ where
...
+        let block_available_gas = self.evm.block().gas_limit - self.gas_used;
+        if tx.tx().gas_limit() > block_available_gas && (self.is_regolith || !is_deposit) {
+            return Err(BlockValidationError::TransactionGasLimitMoreThanAvailableBlockGas {
+                transaction_gas_limit: tx.tx().gas_limit(),
+                block_available_gas,
+            }
+            .into());
+        }
+
+        let hash = tx.tx().trie_hash();
```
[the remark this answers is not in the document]
yongkangc: thanks for the spot, will move that up :)

crates/evm/src/block/mod.rs
```diff
@@ -195,7 +195,55 @@ pub trait BlockExecutor {
...
+    /// and applies the state changes, updates gas accounting, and generates a receipt.
+    ///
+    /// Returns the gas used by the transaction.
+    ///
+    /// # Parameters
+    /// - `output`: The transaction output containing execution result and state changes
+    /// - `tx`: The original transaction (needed for receipt generation)
+    fn commit_transaction(
+        &mut self,
+        output: revm::context_interface::result::ResultAndState<<Self::Evm as Evm>::HaltReason>,
```
klkvr, line 244: nit: let's import the `ResultAndState` instead and use it here instead of fully-qualified path

crates/op-evm/src/block/mod.rs
```diff
@@ -123,15 +119,22 @@ where
...
+    // Deposit transactions require capturing the depositor's nonce BEFORE execution (as execution
+    // increments it), but using it AFTER execution in the receipt. The decomposed API doesn't
+    // provide a clean way to pass this pre-execution state between methods without introducing
+    // stateful coupling, so we keep the monolithic implementation for OP chains.
     fn execute_transaction_with_commit_condition(
         &mut self,
         tx: impl ExecutableTx<Self>,
-        f: impl FnOnce(&ExecutionResult<<Self::Evm as Evm>::HaltReason>) -> CommitChanges,
+        f: impl FnOnce(
+            &revm::context::result::ExecutionResult<<Self::Evm as Evm>::HaltReason>,
```
klkvr, line 131: same here

crates/op-evm/src/block/mod.rs
```diff
@@ -214,6 +217,100 @@ where
...
+        tx: &impl ExecutableTx<Self>,
+    ) -> Result<u64, BlockExecutionError> {
+        let ResultAndState { result, state } = output;
+        let is_deposit = tx.tx().ty() == DEPOSIT_TRANSACTION_TYPE;
+
+        // Cache the depositor account prior to the state transition for the deposit nonce.
+        // Note that this *only* needs to be done post-regolith hardfork, as deposit nonces
+        // were not introduced in Bedrock. In addition, regular transactions don't have deposit
+        // nonces, so we don't need to touch the DB for those.
+        let depositor = (self.is_regolith && is_deposit)
```
klkvr, line 256: i.e I'd expect this to work correctly already

klkvr (changes requested):

crates/op-evm/src/block/mod.rs
klkvr: yeah I'd expect so

crates/op-evm/src/block/mod.rs
```diff
@@ -123,15 +119,19 @@ where
         Ok(())
     }
 
-    fn execute_transaction_with_commit_condition(
+    // No need to override execute_transaction_with_commit_condition anymore.
+    // The default implementation will correctly use our execute_transaction_without_commit
+    // and commit_transaction methods, which properly handle deposit nonces by fetching
+    // them from the DB during the commit phase.
+
```
klkvr, line 126:
```suggestion
```

crates/op-evm/src/block/mod.rs
yongkangc: updated, thanks for that

crates/op-evm/src/block/mod.rs
```diff
@@ -123,15 +119,22 @@ where
         Ok(())
     }
 
+    // OpBlockExecutor must override the default implementation to correctly handle deposit nonces.
+    // Deposit transactions require capturing the depositor's nonce BEFORE execution (as execution
+    // increments it), but using it AFTER execution in the receipt. The decomposed API doesn't
+    // provide a clean way to pass this pre-execution state between methods without introducing
+    // stateful coupling, so we keep the monolithic implementation for OP chains.
```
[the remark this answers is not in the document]
yongkangc: thanks for the clarification @klkvr

yongkangc:
side note - i did a reth bench on op-evm and was successful for 2000 blocks @mattsse @klkvr 

just to test the commit

crates/evm/src/block/mod.rs
```diff
@@ -195,7 +198,52 @@ pub trait BlockExecutor {
...
+    ///
+    /// Returns the gas used by the transaction.
+    ///
+    /// # Parameters
+    /// - `output`: The transaction output containing execution result and state changes
+    /// - `tx`: The original transaction (needed for receipt generation)
+    fn commit_transaction(
+        &mut self,
+        output: ResultAndState<<Self::Evm as Evm>::HaltReason>,
+        tx: &impl ExecutableTx<Self>,
```
mattsse, line 245: I think here we can still keep `impl Executable` and don't need the extra &?

crates/evm/src/block/mod.rs
```diff
@@ -195,7 +198,52 @@ pub trait BlockExecutor {
...
+    /// Returns a [`revm::context_interface::result::ResultAndState`] containing the execution
+    /// result and state changes.
+    ///
+    /// # Use Cases
+    /// - Transaction simulation without affecting state
+    /// - Inspecting transaction effects before committing
+    /// - Building custom commit logic
+    fn execute_transaction_without_commit(
+        &mut self,
+        tx: &impl ExecutableTx<Self>,
```
mattsse, line 228: wonder if we we can actually du `ExecutableTx+ Asref<Self>` here?

mattsse (commented):
last thoughts, pending @klkvr

crates/evm/src/block/mod.rs
yongkangc: @mattsse Did you mean something like `impl ExecutableTx<Self> + AsRef<Transaction>` to allow passing owned values or smart pointers?

would think that The current `&impl ExecutableTx<Self>` already handles borrowed types? might not be 100% understanding the intention for this

yongkangc: makes sense for me

Issue #164, opened 2025-08-26
[Feature] Pass the TxEnv to precompiles as well.


nuntax:
### Describe the feature you would like

I am using the PrecompilesMap for adding and managing Precompiles. However, I now ran into the issue that the EvmInternals field in the PrecompileInput only contains the BlockEnv and not the TxEnv, which I need to access in a precompile.
https://github.com/alloy-rs/evm/blob/785be28496b0e61003536c5f0f740261d9026ce2/crates/evm/src/traits.rs#L158-L161
Would also open a PR myself and add it, just don't know if there's maybe a specific reason that the TxEnv is not in there.

mattsse:
@mablr perhaps you want to take this :) ?

nuntax:
The issue is that the Transaction trait isn't dyn compatible. I have a fork which I use for now in which I solved this by adding a wrapper dyn compatible TransactionTr trait, similar to the EvmInternalsTr trait. It's quite a bit messy though and I didn't really know if there's a better solution.

mablr:
@mattsse I'll have a look by the end of the day 👍

PR #168, opened 2025-08-29
feat: Add TxEnv to EvmInternals

nuntax:
## Motivation

Makes it possible to access the Transaction data in precompiles, closes https://github.com/alloy-rs/evm/issues/164.
## Solution
Ideally, we could just add a Transaction dyn object but since the Transaction trait isn't dyn compatible I added a wrapper trait, TransactionTr. It's missing the access_list and authorization_list method since those use associated types which should be the reason why the Transaction trait isn't dyn compatible in the first place. 

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/evm/src/traits.rs
+++ b/crates/evm/src/traits.rs
@@ -6,7 +6,7 @@ use alloy_primitives::{Address, Bytes, Log, TxKind, B256, U256};
 use core::{error::Error, fmt, fmt::Debug};
 use revm::{
     context::{
-        journaled_state::{account::JournaledAccountTr, TransferError},
+        journaled_state::{account::JournaledAccountTr, JournalCheckpoint, TransferError},
         result::InvalidTransaction,
         Block, Cfg, ContextTr, DBErrorMarker, JournalTr,
     },
@@ -365,6 +365,12 @@ trait EvmInternalsTr: Database<Error = ErasedError> + Debug {
     fn tload(&mut self, address: Address, key: StorageKey) -> StorageValue;
 
     fn tstore(&mut self, address: Address, key: StorageKey, value: StorageValue);
+
+    fn checkpoint(&mut self) -> JournalCheckpoint;
+
+    fn checkpoint_commit(&mut self);
+
+    fn checkpoint_revert(&mut self, checkpoint: JournalCheckpoint);
 }
 
 /// Helper internal struct for implementing [`EvmInternals`].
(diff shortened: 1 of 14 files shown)


mattsse:
cc @mablr

crates/evm/src/traits.rs
```diff
@@ -40,6 +40,193 @@ impl EvmInternalsError {
     }
 }
 
+/// dyn-compatible trait for accessing transaction fields.
+pub trait TransactionTr {
```
mattsse, line 44: I think this is basically alloy_consensus::Transaction, right?

the len and iter calls of the revm traits make this a but tricky -.-
but I dont see another way to do this otherwise

mattsse (commented):
still makes sense imo

although this isnt super nice

wdyt @klkvr

crates/evm/src/traits.rs
nuntax: I didn't think about using alloy_consensus::Transaction tbh. I can't check right now but if it's dyn compatible it might be a better choice for this no? Can take a look later

nuntax: I thought about this some more and a `TxEnv` snapshot in `EvmInternals::new` could be the cleanest path for this.

`new` is already generic over `impl Transaction`, so we can project once at construction time and avoid object-safety workarounds. This keeps access to common tx fields and would remove the TransactionTr trait.

crates/evm/src/traits.rs
```diff
@@ -43,6 +44,203 @@ impl EvmInternalsError {
     }
 }
 
+/// dyn-compatible trait for accessing transaction fields.
```
mattsse: can we document more why we even need to do this

crates/evm/src/traits.rs
```diff
@@ -43,6 +44,203 @@ impl EvmInternalsError {
...
+    fn effective_balance_spending(
+        &self,
+        base_fee: u128,
+        blob_price: u128,
+    ) -> Result<U256, InvalidTransaction>;
+}
+
+impl<T> TransactionTr for T
+where
+    T: Transaction,
```
mattsse: can we use fullpath here to make it obvious which tx trait we mean here

crates/evm/src/traits.rs
```diff
@@ -246,24 +444,27 @@ pub struct EvmInternals<'a> {
...
 
 impl<'a> EvmInternals<'a> {
     /// Creates a new [`EvmInternals`] instance.
-    pub fn new<T>(
+    pub fn new<T, TX>(
         journal: &'a mut T,
         block_env: &'a dyn Block,
         cfg_env: &'a impl Cfg,
-        tx_env: &'a impl Transaction,
+        tx_env: &'a TX,
```
mattsse: I assume we should do dyn TransactionTr here to mirror

block_env: &'a dyn Block,

?

mattsse (approved):
lgtm

pending @klkvr

Issue #170, opened 2025-09-11
[Feature] Introduce `EvmEnv` construction helpers


klkvr:
### Describe the feature you would like

Those functions can be useful outside of reth https://github.com/paradigmxyz/reth/blob/a28c22db39426b7c88bce84f2c0d2ea5862bde08/crates/ethereum/evm/src/lib.rs#L157-L266

We can start by introducing a helpers like `fn evm_env_for_block(header: &Header, blob_params: &BlobParams, chain_id: ChainId)` and evaluate whether it is possible to also reduce code duplication between all of `evm_env`/`next_evm_env`/`evm_env_for_payload`

PR #173, opened 2025-09-15
feat: Add constructor of `EvmEnv` for a block

RomanHodulak:
Part of #170 

## Motivation

There is some consistency rules that apply at all times that are missing in this repository.

## Solution

Add a constructor for `EvmEnv` that takes block `Header` along with other necessary attributes.

Alternative with traits https://github.com/alloy-rs/evm/pull/177

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Breaking changes

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,7 +41,7 @@ alloy-op-evm = { version = "0.21.1", path = "crates/op-evm", default-features = 
 # alloy
 alloy-eip2124 = { version = "0.2", default-features = false }
 alloy-chains = { version = "0.2.0", default-features = false }
-alloy-eips = { version = "1.0", default-features = false }
+alloy-eips = { version = "1.0.34", default-features = false }
 alloy-consensus = { version = "1.0.27", default-features = false }
 alloy-primitives = { version = "1.0.0", default-features = false }
 alloy-sol-types = { version = "1.0.0", default-features = false }


crates/evm/src/env.rs
```diff
@@ -26,6 +30,61 @@ impl<Spec> EvmEnv<Spec> {
...
+        header: &impl BlockHeader,
+        chain_id: ChainId,
+        spec: Spec,
+        blob_params: Option<BlobParams>,
+    ) -> Self
+    where
+        Spec: EthereumHardforks,
+    {
+        /// Maximum transaction gas limit as defined by [EIP-7825](https://eips.ethereum.org/EIPS/eip-7825) activated in `Osaka` hardfork.
+        const MAX_TX_GAS_LIMIT_OSAKA: u64 = 2u64.pow(24);
```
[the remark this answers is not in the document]
mattsse: yep makes sense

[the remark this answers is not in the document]
RomanHodulak: https://github.com/alloy-rs/alloy/pull/2906

crates/op-evm/src/env.rs
```diff
@@ -0,0 +1,24 @@
...
+/// * `chain_id` - The chain identifier.
+/// * `blob_params` - Optional parameters that sets limits on gas and count for blobs.
+pub fn evm_env_for_op_block(
+    header: impl BlockHeader,
+    chain_spec: impl OpHardforks,
+    chain_id: ChainId,
+    blob_params: Option<BlobParams>,
+) -> EvmEnv<OpSpecId> {
+    EvmEnv::for_block(header, chain_spec, chain_id, blob_params, |c, h| crate::spec(c, h))
+}
```
mattsse, line 24: should we also make this a function on `EmvEnv`, because we already have an op feature in that crate

https://github.com/alloy-rs/evm/blob/028ede903fd98fc827afb6d5109a50b604dad9ca/crates/evm/Cargo.toml#L55-L55

crates/evm/src/env.rs
```diff
@@ -1,20 +1,105 @@
...
+    /// * `chain_id` - The chain identifier.
+    /// * `blob_params` - Optional parameters that sets limits on gas and count for blobs.
+    /// * `spec_factory` - A function that derives `Spec` from `chain_spec` and `header`.
+    pub fn for_block<H, C>(
+        header: H,
+        chain_spec: C,
+        chain_id: ChainId,
+        blob_params: Option<BlobParams>,
+        spec_factory: impl FnOnce(&C, &H) -> Spec,
+    ) -> Self
```
mattsse, line 59: I don't think we can generalize this really

because the op init is slightly different:

https://github.com/paradigmxyz/reth/blob/2ac8b7fbd8fd69967904fce430329326d0574c28/crates/optimism/evm/src/lib.rs#L162-L162

I'd rather have dedicated function for optimism here

mattsse (changes requested):

crates/op-evm/src/env.rs
klkvr: yeah I think we could define this under op feature, just need to pull alloy-op-hardforks

crates/evm/src/eth/env.rs
```diff
@@ -0,0 +1,119 @@
...
+    /// * `chain_id` - The chain identifier.
+    /// * `blob_params` - Optional parameters that sets limits on gas and count for blobs.
+    pub fn for_eth_block(
+        header: impl BlockHeader,
+        chain_spec: impl EthereumHardforks,
+        chain_id: ChainId,
+        blob_params: Option<BlobParams>,
+    ) -> Self {
+        /// Maximum transaction gas limit as defined by [EIP-7825](https://eips.ethereum.org/EIPS/eip-7825) activated in `Osaka` hardfork.
+        const MAX_TX_GAS_LIMIT_OSAKA: u64 = 2u64.pow(24);
```
klkvr, line 29: this is released in alloy 1.0.34 now

PR #174, opened 2025-09-16
feat: Add `SpecId` mapping methods

RomanHodulak:
## Motivation

There is a missing link between `SpecId`, which identifies an EVM spec and an Ethereum Hardfork.

## Solution

Add a method that maps Hardfork to SpecId.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/evm/src/spec.rs
```diff
@@ -0,0 +1,216 @@
...
+///   activation.
+///
+/// It is not intended for this trait to be implemented manually. Instead, any [`EthereumHardforks`]
+/// gets `SpecMap` implementation automatically thanks to a blanket implementation.
+///
+/// [`spec_for_header()`]: Self::spec_for_header
+/// [`spec_at_timestamp()`]: Self::spec_at_timestamp
+/// [`spec_at_height()`]: Self::spec_at_height
+/// [`spec_at()`]: Self::spec_at
+pub trait SpecMap {
```
mattsse, line 23: imo we don't really need this extension trait

from caller pov, it's not a big deal if you call `spec::spec_for_header` instead `chain.spec_for_header`

I also want to avoid conflicts with the op variant because the OpHardforks is also EthereumHardforks, which would have conflicting defaults here

mattsse (changes requested):
can we do the same for op as well here?

and keep the functions as is

mattsse (approved):
nice just `spec` is a lot cleaner

PR #177, opened 2025-09-19
feat: Add constructor of `EvmEnv` for a block using traits

RomanHodulak:
## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/evm/src/op/env.rs
```diff
@@ -33,25 +37,46 @@ impl EvmEnv<OpSpecId> {
...
+        /// * `chain_spec` - The chain hardfork description, must implement [`OpHardforks`].
+        /// * `chain_id` - The chain identifier.
+        /// * `blob_params` - Optional parameters that sets limits on gas and count for blobs.
+        pub fn for_op_payload(
+            payload: ExecutionPayload,
+            chain_spec: impl OpHardforks,
+            chain_id: ChainId,
+        ) -> Self {
+            Self::for_op(EvmEnvInput::from_payload(payload), chain_spec, chain_id)
+        }
```
klkvr, line 80: we need to use `OpExecutionData` here https://github.com/alloy-rs/op-alloy/blob/e0db0e171846f72b0c77f24528d72feab594d193/crates/rpc-types-engine/src/envelope.rs#L109

klkvr (changes requested):

crates/evm/src/op/env.rs
RomanHodulak: Thanks for the correction.

I put `OpExecutionPayload` because `OpExecutionData` contains the payload and a sidecar which is not needed here. Wdyt?

PR #182, opened 2025-09-23
feat: Add next block constructors for `EvmEnv`

RomanHodulak:
Part of #170 

## Motivation

The `EvmEnv` is missing constructors for building next block, which is made out of a parent header and some attributes that normally come from block header, but there is no block header yet.

## Solution

Add `for_eth_next_block` and `for_op_next_block` constructors for `EvmEnv`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/evm/src/eth/env.rs
+++ b/crates/evm/src/eth/env.rs
@@ -41,6 +41,7 @@ impl EvmEnv<SpecId> {
     /// * `blob_params` - Optional parameters that sets limits on gas and count for blobs.
     pub fn for_eth_next_block(
         header: impl BlockHeader,
+        attributes: NextEvmEnvAttributes,
         base_fee_per_gas: u64,
         elasticity_multiplier: u128,
         chain_spec: impl EthereumHardforks,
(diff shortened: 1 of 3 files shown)


crates/evm/src/eth/env.rs
```diff
@@ -72,6 +105,78 @@ impl EvmEnv<SpecId> {
...
+            .transitions_at_block(input.height)
+        {
+            let elasticity_multiplier = input.elasticity_multiplier.unwrap_or_default();
+
+            // multiply the gas limit by the elasticity multiplier
+            gas_limit *= elasticity_multiplier as u64;
+
+            // set the base fee to the initial base fee from the EIP-1559 spec
+            basefee = INITIAL_BASE_FEE
+        }
```
klkvr, line 163: it's likely that this logic is never actually entered now so we can explore removing it along with `elasticity_multiplier` arg separately

klkvr (approved):

PR #184, opened 2025-09-25
refactor: Remove London boundary hardfork branch in `EvmEnv` constructor for next block

RomanHodulak:
## Motivation

As per @klkvr suggestion, this branch will not be reached for new blocks.

## Solution

Remove the branch and `elasticity_multiplier` attribute.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mattsse (approved):
this makes sense because we dont expect this to be used for pre london blocks because this is intended for block building

PR #202, opened 2025-10-23
fix: blob fee calc

klkvr:
## Motivation

Unifies `for_eth` and `for_eth_next`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/evm/src/eth/env.rs
+++ b/crates/evm/src/eth/env.rs
@@ -135,7 +135,8 @@ impl EvmEnvInput {
             mix_hash: Some(attributes.prev_randao),
             difficulty: U256::ZERO,
             gas_limit: attributes.gas_limit,
-            // If header does not have blob fields, assume that excess blob gas is 0.
+            // If header does not have blob fields, but we have blob params, assume that excess blob
+            // gas is 0.
             excess_blob_gas: parent
                 .maybe_next_block_excess_blob_gas(blob_params)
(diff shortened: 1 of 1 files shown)


crates/evm/src/eth/env.rs
```diff
@@ -95,99 +95,52 @@ impl EvmEnv<SpecId> {
...
+        blob_params: Option<BlobParams>,
     ) -> Self {
         Self {
             timestamp: attributes.timestamp,
-            height: header.number() + 1,
+            number: parent.number() + 1,
             beneficiary: attributes.suggested_fee_recipient,
             mix_hash: Some(attributes.prev_randao),
-            difficulty: header.difficulty(),
+            difficulty: U256::ZERO,
```
mattsse: makes sense

crates/evm/src/eth/env.rs
```diff
@@ -95,99 +95,52 @@ impl EvmEnv<SpecId> {
...
             gas_limit: attributes.gas_limit,
-            excess_blob_gas: header.excess_blob_gas(),
+            // If header does not have blob fields, but we have blob params, assume that excess blob
+            // gas is 0.
```
mattsse: right, because this starts out at 0

mattsse (approved):

PR #220, opened 2025-11-08
feat(evm): moving rpc conversion traits from Reth

### LINKED ISSUE #219: [Feature] Move reth conversion traits to alloy-evm

we have these helpers:

https://github.com/paradigmxyz/reth/blob/7997cd42832f4be41e0a20e5796f8dc194b7b137/crates/rpc/rpc-convert/src/transaction.rs#L521-L521

which provide common logic, e.g. we have the same thing in foundry, I think twice

## TODO
* move trait def and impls to evm, we can introduce a rpc feature like 
https://github.com/alloy-rs/evm/blob/8a79e5398c7000a8b84d10e2780f9f9c0a3e8b13/crates/evm/Cargo.toml#L61-L61

### Additional context

@mablr

mablr:
## Motivation

Close #219 

## Solution

- Moved trait/structs/impls from `reth-rpc-convert` crate:
  - Fee validation and calculation logic (`CallFees`, `CallFeesError`)
  - Transaction environment conversion (`TryIntoTxEnv`)
- Introduced `rpc`/`op-rpc` features
- The OP’s impls are placed in the alloy-evm’s `op` module because it’s not possible to implement a trait for a struct that are both defined in other crates.

## PR Checklist

- [x] Added Tests (copied from Reth, maybe more needed)
- [x] Added Documentation (copied and updated from Reth)
- [ ] Breaking changes

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,8 +56,8 @@ op-alloy-rpc-types = { version = "0.22", default-features = false }
 op-alloy-rpc-types-engine = { version = "0.22", default-features = false }
 
 # revm
-revm = { version = "31.0.0", default-features = false }
-op-revm = { version = "12.0.0", default-features = false }
+revm = { version = "32.0.0", default-features = false }
+op-revm = { version = "13.0.0", default-features = false }
 
 # misc
 auto_impl = "1"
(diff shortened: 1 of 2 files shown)


crates/evm/src/op/rpc.rs
mattsse: grouping this here under op makes sense to me

mattsse (changes requested):
this is great!

could you perhaps prepare a reth pr with patched alloy-evm deps that replace the reth deps, this way we can figure out if we missed anything or need something else

crates/evm/src/op/rpc.rs
mattsse: yeah perhaps we can just combine them and feature gate op entirely

mattsse (approved):
lgtm

pending @klkvr

PR #221, opened 2025-11-10
feat: bumps nonce of account

rakita:
Previous function `bump_nonce_journal_entry` would just add entry to the journal, it will not bump the nonce.

As usage of this function is not what would be expected, i renamed it to `bump_none`.


crates/evm/src/traits.rs
```diff
@@ -68,7 +68,7 @@ trait EvmInternalsTr: Database<Error = ErasedError> + Debug {
     /// Increments the nonce of the account.
     ///
     /// This creates a new journal entry with this change.
```
mattsse: is this still correct?

mattsse (approved):
lgtm

pending @klkvr

PR #238, opened 2025-12-04
feat: Clean up StateDB methods, bump revm

0xForerunner:
This PR removes unnecessary methods from `StateDB` in favour of using `DatabaseCommitExt`. See #234 and [this pr](https://github.com/bluealloy/revm/pull/3205) for more details.

Additionally we've added a few more trait methods to abstract away from `revm::State`.

You can see why this is required by taking a look at [this pr into reth](https://github.com/paradigmxyz/reth/pull/22098).
## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Breaking changes

### DIFF
--- a/crates/evm/src/block/state.rs
+++ b/crates/evm/src/block/state.rs
@@ -1,5 +1,6 @@
 //! State database abstraction.
 
+use core::fmt::Debug;
 use revm::{
     database::{states::bundle_state::BundleRetention, BundleState, State},
     DatabaseCommit,
@@ -40,7 +41,7 @@ pub trait StateDB: Database + DatabaseCommit {
     fn merge_transitions(&mut self, retention: BundleRetention);
 }
 
-impl<DB: revm::Database + std::fmt::Debug> StateDB for State<DB> {
+impl<DB: revm::Database + Debug> StateDB for State<DB> {
     fn set_state_clear_flag(&mut self, has_state_clear: bool) {
         Self::set_state_clear_flag(self, has_state_clear);
     }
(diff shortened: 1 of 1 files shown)


0xForerunner:
Apologies for the delay @mattsse @klkvr, looks like I forgot about this for a while. Should be ready to go as everything has been merged in `revm`

0xForerunner:
@klkvr just rebased on main. Should be good to go!

klkvr:
@0xForerunner it's still not obvious to me why we need those methods to be honest. does your BAL implementation require the block builder to hold your custom database?

0xForerunner:
@klkvr rebased. Should be ready again.

crates/evm/src/block/state.rs
```diff
@@ -1,9 +1,53 @@
...
+        core::mem::take(self.bundle_state_mut())
+    }
+
+    /// Take all transitions and merge them inside [`BundleState`].
+    ///
+    /// This action will create final post state and all reverts so that
+    /// we at any time revert state of bundle to the state before transition
+    /// is applied.
+    fn merge_transitions(&mut self, retention: BundleRetention);
+}
```
klkvr: it is not entirely clear to me why this abstraction is needed, what is your usecase? right now basic block executors are already agnostic of concrete db

0xForerunner: We need this abstraction because our access list block builder needs to be able to wrap `revm::State`

0xForerunner: I'm not entirely sure what you're suggesting. [We already have StateDB used in the BlockExecutor impls](https://github.com/alloy-rs/evm/blob/main/crates/evm/src/eth/block.rs#L116-L116). What is your reasoning for not wanting this? The previous implementation relied directly on revm::State, which is strictly less flexible.

klkvr: yep and current implementation works with any database which is very flexible which is why I'm a bit hesitant on making it stricter

PR #240, opened 2025-12-07
wip:feat: bal aware block execution

mattsse:
wip scaffolding for validation (block execution) with bal context


mattsse (commented):
this is now revm native

PR #256, opened 2026-01-17
feat(tx): enable zero-copy IntoTxEnv implementations

gakonst:
## Summary

This PR removes the blanket impl `IntoTxEnv for T where T: ToTxEnv` and replaces it with a more targeted blanket impl `IntoTxEnv for &T where T: ToTxEnv`.

## Motivation

The previous blanket impl prevented downstream crates from implementing `IntoTxEnv` directly for owned types. For example, in reth we have a `WithTxEnv<TxEnv, T>` wrapper that contains a `TxEnv`. With the old design, calling `into_tx_env()` on this type would go through `ToTxEnv` which clones the `TxEnv`, even though we could just extract and return it by value.

## Changes

- **Remove** blanket `impl<T, TxEnv> IntoTxEnv<TxEnv> for T where T: ToTxEnv<TxEnv>`
- **Add** blanket `impl<T, Env> IntoTxEnv<Env> for &T where T: ToTxEnv<Env> + ?Sized`
- **Add** explicit `IntoTxEnv` impls for:
  - `Recovered<T>`
  - `WithEncoded<Recovered<T>>`
  - `Either<L, R>`
- **Update** documentation to explain the trait hierarchy and zero-copy pattern

## Usage

Downstream crates can now implement `IntoTxEnv` directly for zero-copy consumption:

```rust
struct WithTxEnv<TxEnv, T> {
    tx_env: TxEnv,
    inner: T,
}

impl<TxEnv, T> IntoTxEnv<TxEnv> for WithTxEnv<TxEnv, T> {
    fn into_tx_env(self) -> TxEnv {
        self.tx_env  // Zero-copy: just moves the inner TxEnv
    }
}
```

References still work via the new blanket impl, so `&MyType` will use `ToTxEnv` (cloning) as before.

### DIFF
--- a/crates/evm/src/tx.rs
+++ b/crates/evm/src/tx.rs
@@ -3,6 +3,26 @@
 //! This module provides traits and implementations for converting various transaction formats
 //! into a unified transaction environment ([`TxEnv`]) that the EVM can execute. The main purpose
 //! of these traits is to enable flexible transaction input while maintaining type safety.
+//!
+//! # Trait Hierarchy
+//!
+//! There are two conversion traits:
+//!
+//! - [`IntoTxEnv`]: Consumes `self` to produce a `TxEnv`. Types that already contain a `TxEnv` (or
+//!   can produce one without cloning) should implement this directly for zero-copy consumption.
+//!
+//! - [`ToTxEnv`]: Builds a `TxEnv` by borrowing from `&self`, typically requiring clones. This is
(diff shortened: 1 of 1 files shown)


Rjected (approved):
Makes sense to have two traits for this if a type can return w/o ownership

mattsse (changes requested):

mattsse (commented):
superseded by #262

PR #260, opened 2026-01-19
feat(block): zero-copy transaction execution via TxParts

mattsse:
## Summary

Introduces zero-copy transaction execution optimization for the BlockExecutor trait via a new `TxParts` abstraction.

### New Types in `crates/evm/src/tx.rs`

- **`TxParts<E, R>`**: Container holding both the EVM transaction environment (`tx_env`) and the recovered transaction data (`recovered`). Enables the `tx_env` to be consumed by the EVM while retaining access to the original transaction for receipt generation.

- **`IntoTxParts`**: Trait for types that can be decomposed into `TxParts`. Implemented for:
  - `Recovered<T>` - builds `TxEnv` via `FromRecoveredTx`
  - `WithEncoded<Recovered<T>>` - builds `TxEnv` via `FromTxWithEncoded`
  - References - clones `TxEnv`, borrows for `RecoveredTx`
  - `TxParts` itself - just moves the fields

### BlockExecutor Changes in `crates/evm/src/block/mod.rs`

- **`ExecutableTx`**: Now requires `IntoTxParts` instead of just `ToTxEnv + RecoveredTx`

- **`execute_transaction_with_commit_condition`**: Refactored default implementation to:
  1. Call `tx.into_tx_parts()` to split the transaction
  2. Call `execute_transaction_without_commit_with_parts(tx_env, &recovered)` - consumes `tx_env` by value
  3. Call `commit_transaction(output, recovered)`

- **`execute_transaction_without_commit_with_parts`**: New method that takes `tx_env` by value for zero-copy consumption

- **`commit_transaction`**: Now accepts `impl RecoveredTx` instead of `impl ExecutableTx`

### EthBlockExecutor Changes

- Implemented `execute_transaction_without_commit_with_parts` that consumes `tx_env` directly

### OpBlockExecutor Changes

- Implemented `execute_transaction_without_commit_with_parts`
- Added `pending_tx_da_footprint` field to store DA footprint computed during execution
- DA footprint is computed once in execution, applied during commit

## Motivation

This enables zero-copy `TxEnv` consumption for `transact()` calls, particularly beneficial for types like `WithTxEnv<TxEnv, Arc<T>>` that pre-compute the transaction environment.

## Testing

- Existing tests pass
- Added compile-time tests for `IntoTxParts` implementations

### DIFF
--- a/crates/evm/src/block/mod.rs
+++ b/crates/evm/src/block/mod.rs
@@ -242,11 +242,11 @@ pub trait BlockExecutor {
     ///
     /// # Parameters
     /// - `tx_env`: The transaction environment to execute (consumed)
-    /// - `tx`: Reference to recovered transaction for validation (gas limit checks)
+    /// - `tx`: Recovered transaction for validation (gas limit checks)
     fn execute_transaction_without_commit_with_parts(
         &mut self,
         tx_env: <Self::Evm as Evm>::Tx,
-        tx: &impl RecoveredTx<Self::Transaction>,
+        tx: impl RecoveredTx<Self::Transaction>,
     ) -> Result<ResultAndState<<Self::Evm as Evm>::HaltReason>, BlockExecutionError>;
 
     /// Executes a single transaction without committing state changes.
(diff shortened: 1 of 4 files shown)


mattsse (commented):
superseded by #262

PR #265, opened 2026-01-23
EIP-7778: add new gas_spent field in receipts

jenpaff:
EIP-7778 introduces a new gas_spent field in receipts to track per-transaction gas after refunds (what users actually pay). This separates it from cumulative_gas_used, which now represents gas before refunds for block limit enforcement. This PR adds the executor-side tracking needed to compute gross gas by accumulating refunds in EthBlockExecutor and adjusting cumulative_gas_used to include them when Amsterdam is active.


mattsse (commented):
this is outdated now

PR #286, opened 2026-02-13
tip1016: revm state-gas integration

rakita:
## Summary
- Update all 13 revm sub-crate revs to `e3223d5b38f7affb901601ed917fa7cdc466be74` (rakita/state-gas branch) for TIP-1016 dual-limit gas accounting

## Test plan
- [x] `cargo check` compiles clean
- [x] `cargo +nightly clippy` passes (only deprecation warning for `record_cost`)
- [x] `cargo nextest run` — all 69 tests pass


klkvr (commented):
looks like this has some conflicts

crates/evm/src/precompiles.rs
```diff
@@ -880,7 +977,15 @@ where
 
 impl From<(PrecompileId, PrecompileFn)> for DynPrecompile {
     fn from((id, f): (PrecompileId, PrecompileFn)) -> Self {
-        let p = move |input: PrecompileInput<'_>| f(input.data, input.gas);
+        let p = move |input: PrecompileInput<'_>| -> PrecompileResultExt {
+            match f(input.data, input.gas) {
+                Ok(output) => Ok(PrecompileOutputExt::from_precompile_output(output, input.gas, 0)),
```
fgimenez: why are we passing 0 here instead of `input.reservoir`?

crates/evm/src/block/mod.rs
```diff
@@ -27,6 +27,56 @@ pub use state::*;
...
+
+impl From<u64> for GasOutput {
+    fn from(gas: u64) -> Self {
+        Self::new(gas)
+    }
+}
+
+impl From<GasOutput> for u64 {
+    fn from(gas: GasOutput) -> Self {
+        gas.tx_gas_used()
```
fgimenez: here we are droppinng `state_gas_used`, is that correct? not used anywhere, but having this impl here could be used in the future

rakita: Should remove it

crates/evm/src/eth/block.rs
```diff
@@ -1,5 +1,7 @@
 //! Ethereum block executor.
 
+use std::cmp::min;
```
fgimenez: otherwise `no_std` build is broken https://github.com/tempoxyz/tempo/actions/runs/24129073435/job/70400870007?pr=3287
```suggestion
use core::cmp::min;
```

crates/evm/src/precompiles.rs
rakita: It is for test purposes

crates/evm/src/block/mod.rs
```diff
@@ -25,6 +25,44 @@ pub use state::*;
...
 
+/// Gas used by a transaction, split into regular and state gas components.
+///
+/// EIP-8037 introduces dual-limit gas accounting with a separate state gas reservation
+/// that tracks gas spent on state creation operations (SSTORE, CREATE, account creation, code
+/// deposit).
+///
+/// - State gas: State-specific gas tracking (storage and contract creation)
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub struct GasOutput {
```
mattsse: this does not belong top of file here

crates/evm/src/eth/block.rs
```diff
@@ -92,23 +99,36 @@ impl<H, T> TxResult for EthTxResult<H, T> {
...
 impl<'a, Evm, Spec, R> EthBlockExecutor<'a, Evm, Spec, R>
 where
-    Spec: Clone,
     R: ReceiptBuilder,
 {
     /// Creates a new [`EthBlockExecutor`]
-    pub fn new(evm: Evm, ctx: EthBlockExecutionCtx<'a>, spec: Spec, receipt_builder: R) -> Self {
+    pub fn new(evm: Evm, ctx: EthBlockExecutionCtx<'a>, spec: Spec, receipt_builder: R) -> Self
+    where
+        Spec: Clone,
```
mattsse, line 107: why change this

crates/evm/src/precompiles.rs
```diff
@@ -540,69 +543,27 @@ where
...
-                } else {
-                    InstructionResult::PrecompileError
-                };
-            }
-        };
+        }
+        .map_err(|e| e.to_string())?;
 
-        Ok(Some(result))
+        Ok(Some(precompile_output_to_interpreter_result(precompile_output, inputs.gas_limit)))
```
mattsse, line 566: why does this exist as a standalone fn and not as a fn of precompile_output or as a constructor?

crates/evm/src/precompiles.rs
```diff
@@ -540,69 +543,27 @@ where
...
-            Err(e) => {
-                result.result = if e.is_oog() {
-                    InstructionResult::PrecompileOOG
-                } else {
-                    InstructionResult::PrecompileError
-                };
-            }
-        };
+        }
+        .map_err(|e| e.to_string())?;
```
mattsse, line 564: why do we need to map this to a string now?

crates/evm/src/eth/block.rs
rakita: Small refactor, nothing more. Clone from line 95 was moved here so `max_block_gas_used` can be in the same `EthBlockExecutor` block

crates/evm/src/precompiles.rs
rakita: `PrecompileOutput` and `InterpreterResults` live in separate crates that do not include each other.

crates/evm/src/block/mod.rs
rakita: Will move it to saparate file

crates/evm/src/precompiles.rs
rakita: `PrecompileError` can be [AnyError](https://github.com/bluealloy/revm/blob/31bee691659382520449de95a03ecd69abf71825/crates/precompile/src/interface.rs#L536) (and Arc over Error) so to not change the return that is String we need to do `to_string`.

crates/evm/src/block/mod.rs
rakita: commit: https://github.com/alloy-rs/evm/pull/286/changes/eef6aeb82213d0eaacfc09917d7741f6b984a7b9

crates/evm/src/block/error.rs
```diff
@@ -79,6 +79,11 @@ pub enum BlockValidationError {
     /// [EIP-6110]: https://eips.ethereum.org/EIPS/eip-6110
     #[error("failed to decode deposit requests from receipts: {_0}")]
     DepositRequestDecode(String),
+    /// Error when block's total gas used exceeds the block gas limit
+    ///
+    /// [EIP-8037]: https://eips.ethereum.org/EIPS/eip-8037
+    #[error("block gas used  exceeds block gas limit")]
```
mattsse:
```suggestion
    #[error("block gas used exceeds block gas limit")]
```

mattsse (dismissed):

mattsse (approved):

crates/evm/src/eth/block.rs
```diff
@@ -161,16 +198,34 @@ where
...
+        // check whether amsterdam is active
+        let amsterdam_active = self
+            .spec
+            .is_amsterdam_active_at_timestamp(self.evm.block().timestamp().saturating_to());
+
+        // Check block gas limit after each transaction if Amsterdam is active, as required by
+        // EIP-8037.
+        if amsterdam_active && self.max_block_gas_used() > self.evm.block().gas_limit() {
+            return Err(BlockValidationError::BlockGasExceeded.into());
+        }
```
klkvr: it is still not clear to me why do we perform this check here. it seems to be a subject to DOS if it fails only before we've executed the transactioj

crates/evm/src/precompiles.rs
klkvr: let's change return to `PrecompileError` on revm side?

not a big deal tho because this should never really happen

rakita: > let's change return to PrecompileError on revm side?

This is good idea

crates/evm/src/eth/block.rs
rakita: We decided to move this check to devnet3 branch, as EIP-8037 is going to be updated in future

klkvr (approved):

PR #304, opened 2026-03-01
feat: add TxEnvironment and CfgEnvironment traits

figtracer:
Add TxEnvironment and CfgEnvironment traits mirroring the existing BlockEnvironment pattern, allowing network-specific transaction and config types to expose their inner revm types via mutable accessors.

- TxEnvironment: wraps revm::context::TxEnv with tx_env_mut()
- CfgEnvironment: wraps revm::context::CfgEnv with cfg_env_mut(), reusing the Spec associated type from the Cfg supertraitx

### DIFF
--- a/crates/evm/src/env.rs
+++ b/crates/evm/src/env.rs
@@ -4,7 +4,7 @@ use core::fmt::Debug;
 
 use alloy_primitives::U256;
 use revm::{
-    context::{BlockEnv, CfgEnv},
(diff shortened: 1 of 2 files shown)


figtracer:
update: this one is actually not that important as thought of initially; there's a much simpler way to approach this than to add these. this adds a bit of indirection but it's just better documented

PR #313, opened 2026-03-13
feat: add `TransactionEnv` trait

decofe:
## Summary

Migrates `TransactionEnv` from `reth_evm` to `alloy-evm`, making it available to downstream consumers.

## Motivation

`TransactionEnv` provides setters (`set_gas_limit`, `set_nonce`, `set_access_list`) on top of `revm::context_interface::Transaction`. These are needed for gas estimation, access list creation, and RPC simulate endpoints. Currently this trait lives in `reth_evm` but has no reth-specific dependencies, so it belongs in `alloy-evm`.

## Changes

- Added `TransactionEnv` trait to `alloy_evm::tx` with `set_gas_limit`, `set_nonce`, `set_access_list` (plus `with_*` builder variants)
- Implemented `TransactionEnv` for `revm::context::TxEnv`, including tx type upgrade from legacy to eip2930 when access list is set

## Testing

`cargo fmt --all --check && cargo clippy --workspace --all-targets -- -D warnings && cargo test` all pass.

Co-Authored-By: Arsenii Kulikov <62447812+klkvr@users.noreply.github.com>

Prompted by: klkvr

### DIFF
--- a/crates/evm/src/tx.rs
+++ b/crates/evm/src/tx.rs
@@ -10,11 +10,13 @@ use alloy_consensus::{
     TxEip4844, TxEip4844Variant, TxEip7702, TxLegacy,
 };
 use alloy_eips::{
-    eip2718::WithEncoded,
+    eip2718::{WithEncoded, EIP2930_TX_TYPE_ID, LEGACY_TX_TYPE_ID},
+    eip2930::AccessList,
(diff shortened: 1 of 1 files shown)


mattsse (approved):
makes sense

PR #314, opened 2026-03-16
feat: integrate EIP-8037 state gas and GasOutput from revm

rakita:
## Summary

This PR integrates EIP-8037 (state gas / reservoir model) and the `GasOutput` struct from:
- **revm commit 556602bc**: Full EIP-8037 state gas implementation
- **alloy-rs/evm PR #286**: GasOutput integration pattern

### 1. **Revm Dependency Update**
- Updated to revm commit `556602bc` which includes:
  - EIP-8037 state gas tracking (SSTORE, CREATE, account creation, code deposit)
  - New `GasTracker` structure for dual-limit gas accounting
  - Spec-based auto-enabling when SpecId >= AMSTERDAM
  - Fixed deprecated `record_cost()` method (now `record_regular_cost()`)

### 2. **Amsterdam Hardfork Support** 
- Added Amsterdam hardfork detection in `spec_id.rs`
- Amsterdam correctly positioned between Osaka and Prague
- Full test coverage

### 3. **GasOutput Integration**
New `GasOutput` struct replacing `u64` gas returns:
```rust
pub struct GasOutput {
    pub regular_gas_used: u64,  // Execution gas
    pub state_gas_used: u64,    // EIP-8037 state creation gas
}
```

Updated trait methods:
- `execute_transaction()` → `Result<GasOutput, BlockExecutionError>`
- `commit_transaction()` → `Result<GasOutput, BlockExecutionError>`

### 4. **EIP-8037 Implementation**
- Gas extraction from revm's `ResultGas::state_gas_spent()`
- Proper tracking through `EthBlockExecutor::commit_transaction()`
- Only non-zero when Amsterdam spec is active

### 5. **API Fixes**
- Updated to new `PrecompileOutput` API (3 parameters)
- Fixed JournalLoadError type handling
- Proper error propagation with `JournalLoadError::unwrap_db_error()`

## Test Results

✅ **All 49 tests passing**
- 26 hardfork mapping tests (including Amsterdam)
- 23 block execution and precompile tests
- No breaking changes

## EIP-8037 Spec Compliance

✅ Dual-limit gas accounting  
✅ Reservoir model with TX_MAX_GAS_LIMIT  
✅ Correct state gas costs  
✅ Proper propagation through nested calls  
✅ Reservoir refill on REVERT/HALT  
✅ GAS opcode returns remaining (excludes reservoir)

## Backward Compatibility

- `GasOutput` implements `From<u64>` for easy migration
- `state_gas_used` defaults to 0 for pre-Amsterdam specs
- All existing code compatible without modification

## Related Issues

- EIP-8037: https://eips.ethereum.org/EIPS/eip-8037
- alloy-rs/evm PR #286
- revm commit 556602bc

---

**Note**: This is a draft PR for review. Ready to merge once feedback is addressed.


crates/evm/src/block/mod.rs
```diff
@@ -275,7 +325,7 @@ pub trait BlockExecutor {
         &mut self,
         tx: impl ExecutableTx<Self>,
         f: impl FnOnce(&ExecutionResult<<Self::Evm as Evm>::HaltReason>),
-    ) -> Result<u64, BlockExecutionError> {
+    ) -> Result<GasOutput, BlockExecutionError> {
```
klkvr, line 328: I think i'm okay with extra type and changed return value here for now, but imo we should consider removing any output from this fn entirely or making it an associated type

there's no strong reason for this to return anything as executor keeps track of the gas internally anyway

crates/evm/src/precompiles.rs
```diff
@@ -614,6 +617,85 @@ where
...
+/// Output of a precompile call.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct PrecompileOutputExt {
+    /// Gas tracker with remaining gas and reservoir info.
+    pub gas: GasTracker,
+    /// Whether the precompile reverted.
+    pub reverted: bool,
+    /// Output bytes.
+    pub bytes: Bytes,
+}
```
klkvr, line 681: we should change `PrecompileOutput` on revm side to contain the `GasTracker` instead

crates/evm/src/precompiles.rs
```diff
@@ -614,6 +617,85 @@ where
...
+pub type PrecompileResultExt = Result<PrecompileOutputExt, PrecompileErrorExt>;
+
+/// Extended precompile error with gas tracking information.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct PrecompileErrorExt {
+    /// The precompile error.
+    pub precompile_error: PrecompileError,
+    /// Gas tracker.
+    pub gas_tracker: GasTracker,
+}
```
klkvr, line 630: what's the motivation for this? isn't every precompile error effectively a halt which is always handled in the same way wrt gas?

crates/evm/src/precompiles.rs
rakita: These extra fields are not used inside revm as it have only simple precomile function, but are used by DynPrecompile in alloy/evm.

I think that `PrecompileOutputExt` should be places where `DynPrecompiles` are. I am okay with moving all of them in revm, wdyt about this?

crates/evm/src/precompiles.rs
rakita: There are `gas_state` and `reservoir` that need to be returned to the parent frame, even in case of halt/revert

The idea was to send full gas_tracker back as it is the main struct that contains all gas related fields.

 Arguably, we could change it to only one field that return's reservoir, wdyt?

crates/evm/src/block/mod.rs
rakita: I think i overegarly used it. I agree with the statement of not returning anything, it makes sense.

rakita: Have testes this, imo we could do it saparetly from this PR it looks like this:
alloy/evm: https://github.com/alloy-rs/evm/pull/326
reth: https://github.com/paradigmxyz/reth/pull/23327

rakita:
Closing it in support of [tip1016](https://github.com/alloy-rs/evm/pull/286) branch

PR #319, opened 2026-03-24
chore: eip 8037 related changes for bal-devnet

Rimeeeeee:
based on pr #314


crates/evm/src/error.rs
```diff
@@ -33,6 +33,12 @@ pub trait InvalidTxError: Error + Send + Sync + Any + 'static {
         })
     }
 
+    /// Returns whether the error is due to the max initcode size being higher than max.
+    fn is_initcode_larger_than_max(&self) -> bool {
+        self.as_invalid_tx_err()
+            .is_some_and(|err| matches!(err, InvalidTransaction::CreateInitCodeSizeLimit))
+    }
+
```
rakita:
```suggestion
```

Should this be removed, not seeing that it is used?

crates/evm/src/eth/receipt_builder.rs
```diff
@@ -19,13 +19,13 @@ pub struct ReceiptBuilderCtx<'a, T, E: Evm> {
...
-    /// After Osaka, `cumulative_gas_used` tracks gas before refunds, and `gas_spent`
-    /// tracks gas after refunds.
-    pub gas_spent: Option<u64>,
+    // Gas spent by the transaction after refunds (what the user pays).
+    //
+    // This is set when EIP-7778 is active (Osaka hardfork).
+    // Before Osaka, this is `None` and `cumulative_gas_used` tracks gas after refunds.
+    // After Osaka, `cumulative_gas_used` tracks gas before refunds, and `gas_spent`
+    // tracks gas after refunds.
+    //pub gas_spent: Option<u64>,
```
rakita:
```suggestion
```

Is this artifact from previous versions? `cumulative_gas_used` should be enought. Do correct me if I am wrong here.

crates/evm/src/error.rs
Rimeeeeee: we dont need this.

crates/evm/src/eth/receipt_builder.rs
Rimeeeeee: yep, this was the devnet 2 changes concerning receipt gas(eip 7778) should be removed

rakita (approved):

PR #323, opened 2026-03-28
Enable system call to be traced by inspector

maoueh:
## Motivation

When using ExEx to replay a full block including the pre/post execution steps, system calls weren't inspected before if inspector was enabled.

## Solution

Just like `transact_raw` does, if the `EthEvm` is in inspect mode, `transact_system_call` now check if `self.inspect` is set and call the inspect variant of 
`inspect_system_call_with_caller`.

### Open Questions

One thing I'm unusure of the impact is about other inspector out there, would they be affected?

Maybe `self.inspect` could be switch off before `transact_system_call` is called in the current code and restored after to keep current behavior.

## PR Checklist

Waiting on initial feedback.

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/evm/src/eth/mod.rs
+++ b/crates/evm/src/eth/mod.rs
@@ -16,7 +16,8 @@ use revm::{
     interpreter::{interpreter::EthInterpreter, InterpreterResult},
     precompile::{PrecompileSpecId, Precompiles},
     primitives::hardfork::SpecId,
-    Context, ExecuteEvm, InspectEvm, Inspector, MainBuilder, MainContext, SystemCallEvm,
+    Context, ExecuteEvm, InspectEvm, InspectSystemCallEvm, Inspector, MainBuilder, MainContext,
+    SystemCallEvm,
 };
 
 mod block;
@@ -232,7 +233,12 @@ where
         contract: Address,
         data: Bytes,
     ) -> Result<ResultAndState<Self::HaltReason>, Self::Error> {
-        self.inner.system_call_with_caller(caller, contract, data)
+        if self.inspect {
+            self.inner
+                .inspect_system_call_with_caller(caller, contract, data)
+        } else {
+            self.inner.system_call_with_caller(caller, contract, data)
+        }
     }
 
     fn finish(self) -> (Self::DB, EvmEnv<Self::Spec>) {


mattsse (commented):
this is a bit tricky, because afaik they dont appear as part of the block level tracing return values, so this could potentially end up including them.

but maybe not if we can make this work in the block level tracer

so I think this change seems fine, but we could perhaps consider adding an additional setting

wdyt @klkvr

maoueh:
> this is a bit tricky, because afaik they dont appear as part of the block level tracing return values, so this could potentially end up including them.

That is my reasoning also, I feel that `debug_traceBlock` could catch those without being aware which could trace some calls it wasn't before.

Geth don't trace system calls via `debug_traceBlock`:

```
	blockCtx := core.NewEVMBlockContext(block.Header(), api.chainContext(ctx), nil)
	evm := vm.NewEVM(blockCtx, statedb, api.backend.ChainConfig(), vm.Config{/* Missing Tracer: tracer here, so no tracing happening with this EVM */})
	if beaconRoot := block.BeaconRoot(); beaconRoot != nil {
		core.ProcessBeaconBlockRoot(*beaconRoot, evm)
	}
	if api.backend.ChainConfig().IsPrague(block.Number(), block.Time()) {
		core.ProcessParentBlockHash(block.ParentHash(), evm)
	}
```

I think this would be an hidden changes.

I'm using `BlockExecutor` via `executor.apply_pre_execution_changes()` so I'm not the direct caller of it. So I'm unsure where this setting should go. Seems like calls sites should take the decision about if they want to trace system calls (pre + post).

> but maybe not if we can make this work in the block level tracer

Would be interested to know more, I saw it but unsure if it's ready to try/improve it. Currently using ExEx than creating a `BlockExecutorFactory` and then tracing full block like that, replicating essentially the tracing we have using Geth Live Tracing feature.

So if `BlockTracer` makes that easier/better, I'm all ears :)

klkvr:
i think we should make sure that any reth rpc callsites explicitly disable inspection before pre/post execution hooks before making this change

maoueh:
@klkvr Ok no problem. I've noted https://github.com/paradigmxyz/reth/blob/26a37f3c00649a0b8f5381b2b9a8b42a312c5d5f/crates/rpc/rpc-eth-api/src/helpers/trace.rs as you main points. Probably others in `reth`.

Is there other repositories I should inspect?

PR #326, opened 2026-04-01
WIPremove GasOutput from BlockExecutor trait methods

rakita:
Investigating


mattsse:
I think we can close this one

PR #330, opened 2026-04-03
chore: new rev for revm bal devnet 3

Rimeeeeee:


rakita (approved):
bump to revm for v5.6.1 fixture

PR #331, opened 2026-04-06
chore: check for gas used at boundary

Rimeeeeee:
* minus `block_regular_gas_used`  from  `block_available_gas`  if amsterdam is enabled.
*  perform boundary checking during commit transaction i.e. ` max(block_regular_gas_used, block_state_gas_used)` donot exceed `block gas limit`.

cc: @rakita

### DIFF
--- a/crates/evm/src/eth/block.rs
+++ b/crates/evm/src/eth/block.rs
@@ -165,13 +165,13 @@ where
         // Pre-Amsterdam: use tx_gas_used (gas after refunds) as cumulative gas, matching
         // the original behavior where gas_used = spent - refunded.
         //
-        // Amsterdam+: use max(block_regular_gas_used, block_state_gas_used) which tracks
+        // Amsterdam+: use block_regular_gas_used which tracks
         // gas without refunds, as required by EIP-8037 dual-limit accounting.
         let block_gas_used = if self
             .spec
             .is_amsterdam_active_at_timestamp(self.evm.block().timestamp().saturating_to())
         {
-            self.max_block_gas_used()
+            self.block_regular_gas_used
         } else {
             self.cumulative_tx_gas_used
         };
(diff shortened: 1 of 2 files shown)


crates/evm/src/eth/block.rs
Rimeeeeee: do we need to do min( tx.tx().gas_limit() ,MAX_TX_LIMIT) here? @rakita

rakita: Not here but on line 174

Rimeeeeee: Gas limit on line 174???

rakita: hm, gas_limit for pre amsteram can't be more than MAX_TX_LIMIT so technically it is okay to do min (tx.tx().gas_limit() ,MAX_TX_LIMIT) but i would saparate this, and be explicit about it

crates/evm/src/eth/block.rs
```diff
@@ -165,13 +165,13 @@ where
...
         //
-        // Amsterdam+: use max(block_regular_gas_used, block_state_gas_used) which tracks
+        // Amsterdam+: use block_regular_gas_used which tracks
         // gas without refunds, as required by EIP-8037 dual-limit accounting.
         let block_gas_used = if self
             .spec
             .is_amsterdam_active_at_timestamp(self.evm.block().timestamp().saturating_to())
         {
-            self.max_block_gas_used()
+            self.block_regular_gas_used
```
rakita, line 178:
```suggestion
            (min(tx_regular_gas,MAX_TX_LIMIT), self.block_regular_gas_used)
```

crates/evm/src/eth/block.rs
rakita: https://github.com/alloy-rs/evm/pull/331/changes#r3039619457

rakita: And check if you can access revm cfg.tx_gas_limit_cap() function

crates/evm/src/eth/block.rs
```diff
@@ -217,6 +226,12 @@ where
         self.block_state_gas_used += state_gas_used;
         self.cumulative_tx_gas_used += tx_gas_used;
 
+        // Check block gas limit after each transaction if Amsterdam is active, as required by
+        // EIP-8037.
+        if amsterdam_active && self.max_block_gas_used() > self.evm.block().gas_limit() {
+            return Err(BlockValidationError::BlockGasExceeded.into());
+        }
```
rakita, line 233: @klkvr is this a good place for this check? After state gas this got moved at after tx exec

PR #337, opened 2026-04-13
chore: bump alloy to 2.0.0

mattsse:
Bump alloy dependencies to the released 2.0.0 versions. No code changes required.

Supersedes #334


mattsse:
Updated #334 directly instead

PR #339, opened 2026-04-15
chore(BAL): allow tracking of zero balance change for withdrawal

Rimeeeeee:


rakita (approved):

mattsse (approved):
context: this causes hive failures for BAL tests, however 0 withdrawals aren't reachable, so this change is fine.

and also:

https://github.com/ethereum/go-ethereum/blob/822e7c648618a43e7f8326868b9334b9976185c8/consensus/beacon/consensus.go#L344-L356

PR #343, opened 2026-04-17
fix: use EVM-configured tx gas limit cap

klkvr:
## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/evm/src/eth/block.rs
+++ b/crates/evm/src/eth/block.rs
@@ -24,7 +24,7 @@ use alloy_hardforks::EthereumHardfork;
 use alloy_primitives::{Bytes, Log, B256};
 use revm::{
     context::Block, context_interface::result::ResultAndState, database::DatabaseCommitExt,
-    primitives::eip7825::TX_GAS_LIMIT_CAP, DatabaseCommit, Inspector,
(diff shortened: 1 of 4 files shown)


crates/evm/src/eth/block.rs
```diff
@@ -177,9 +177,12 @@ where
 
         // Use regular part of transaction gas limit to check if it fits inside available block
         // space.
-        let tx_min_gas_limit = min(tx.tx().gas_limit(), TX_GAS_LIMIT_CAP);
+        let mut max_tx_gas_usage = tx.tx().gas_limit();
+        if let Some(tx_gas_limit_cap) = self.evm.cfg_env().tx_gas_limit_cap {
+            max_tx_gas_usage = min(max_tx_gas_usage, tx_gas_limit_cap);
+        }
```
rakita, line 183:
```suggestion
    let mut max_tx_gas_usage = min(tx.tx().gas_limit(), self.evm.cfg_env().tx_gas_limit_cap());
```

It should be this, if `tx_gas_limit_cap` is none it will set Osaka max gas limit

rakita (approved):

PR #348, opened 2026-04-26
fix: enforce state-gas budget against tx gas limit

rakita:
## Summary
- Add a second per-tx gas-limit check in the Ethereum block executor: on Amsterdam+ blocks, the uncapped `tx.gas_limit` must also fit in the remaining state-gas budget (`block.gas_limit - block_state_gas_used`). Without this, state gas could push the block past its overall gas limit, since `tx_gas_limit_cap` does not bound state gas.
- Introduce a distinct `BlockValidationError::TransactionGasLimitMoreThanAvailableBlockStateGas` variant so callers can tell a state-gas overflow apart from a regular-gas overflow.

## Test plan
- [ ] `cargo check -p alloy-evm` (passes locally)
- [ ] Add an integration test that submits a tx whose `gas_limit` exceeds the remaining state-gas budget and asserts the new error variant is returned
- [ ] Verify pre-Amsterdam blocks are unaffected (new branch is gated on `enable_amsterdam_eip8037`)

### DIFF
--- a/crates/evm/src/block/error.rs
+++ b/crates/evm/src/block/error.rs
@@ -29,6 +29,16 @@ pub enum BlockValidationError {
         /// The available block gas
         block_available_gas: u64,
     },
+    /// Error when transaction gas limit exceeds available block state gas (EIP-8037).
+    #[error(
+        "transaction gas limit {transaction_gas_limit} is more than blocks available state gas {block_available_state_gas}"
+    )]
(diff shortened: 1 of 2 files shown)


rakita:
Added https://github.com/alloy-rs/evm/blob/main/crates/evm/src/eth/block.rs#L222-L235

PR #350, opened 2026-04-28
feat(bal-devnet-4): integrate revm (rev 3ed3bdfe)

rakita:
## Summary
- Bump pinned revm git rev to `3ed3bdfed9ad6e5ba37f4e1f015436ab89ca98be` ("refactor(eip8037): drop block_gas_limit from Cfg::cpsb")
- Bump workspace `revm` version constraint from 37.0.0 → 38.0.0 so the `[patch.crates-io]` entries actually apply (the new rev publishes revm at 38.0.0; without the bump cargo silently falls back to the registry release)

## Test plan
- [x] `cargo build --workspace`
- [x] `cargo check --workspace --all-targets`


rakita:
Obsolete by newer devnet

PR #357, opened 2026-05-11
refactor: introduce `BalEvm`

klkvr:
introduces `BalEvm` which allows `EthBlockExecutor` to drive BAL building/verification internally, and adds BAL to `BlockExecutionResult`

### DIFF
--- a/crates/evm/src/eth/block.rs
+++ b/crates/evm/src/eth/block.rs
@@ -15,15 +15,21 @@ use crate::{
         BlockValidationError, ExecutableTx, GasOutput, OnStateHook, StateChangePostBlockSource,
         StateChangeSource, StateDB, SystemCaller, TxResult,
     },
-    Evm, EvmFactory, FromRecoveredTx, FromTxWithEncoded, RecoveredTx,
+    eth::EthEvmContext,
+    evm::BalEvm,
+    precompiles::PrecompilesMap,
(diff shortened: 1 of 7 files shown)


mattsse (approved):

crates/evm/src/eth/block.rs
```diff
@@ -301,6 +343,14 @@ where
 
         self.evm.db_mut().commit(balance_increment_state);
 
+        let bal = self.evm.take_built_bal();
+
+        if let Some(provided_bal) = &self.ctx.block_access_list {
+            if bal.as_ref() != Some(provided_bal.as_ref()) {
+                return Err(BlockValidationError::msg("invalid BAL provided").into());
```
mattsse, line 362: should we make this a dedicated error?

mattsse (approved):

crates/evm/src/evm.rs
```diff
@@ -190,6 +192,11 @@ pub trait Evm {
         self.components_mut().1
     }
 
+    /// Returns this EVM as a mutable BAL-capable EVM, if supported.
+    fn as_bal_evm_mut(&mut self) -> Option<&mut dyn BalEvm> {
+        None
+    }
```
mattsse, line 198: I like this approach, this should be sufficient enough, and usually we only ever have one implementation of this, if this fails to support bals at runtime this would then always indicate an implemantation issue, imo it's fine to enfrce this at runtime

mattsse (approved):

PR #363, opened 2026-05-21
fix: dispatch proper balance increments state to state hook

klkvr:
## Motivation

Right now the state produced by `balance_increment_state` and dispatched to state hook contains wrong `original_info`. At the moment it doesn't present an issue for reth but it'd be easier to implement some optimizations if this value was right.

This PR fixes the `balance_increment_state` to produce correct state and uses it as both state hook input and the commited value. `balance_increment_state` visibility is changed to avoid misuse as its behavior is slightly different now

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/evm/src/block/state_changes.rs
+++ b/crates/evm/src/block/state_changes.rs
@@ -102,7 +102,6 @@ pub fn insert_post_block_withdrawals_balance_increments(
 
 /// Creates an `EvmState` from a map of balance increments and the current state
 /// to load accounts from. No balance increment is done in the function.
-/// Zero balance increments are ignored and won't create state entries.
 pub(crate) fn balance_increment_state<DB>(
     balance_increments: &AddressMap<u128>,
     state: &mut DB,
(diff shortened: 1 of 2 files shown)


crates/evm/src/block/state_changes.rs
```diff
@@ -102,31 +102,26 @@ pub fn insert_post_block_withdrawals_balance_increments(
 
 /// Creates an `EvmState` from a map of balance increments and the current state
 /// to load accounts from. No balance increment is done in the function.
-/// Zero balance increments are ignored and won't create state entries.
-pub fn balance_increment_state<DB>(
+pub(crate) fn balance_increment_state<DB>(
```
mattsse, line 105: this breaking and it looks like optimism is using this

https://github.com/ethereum-optimism/optimism/blob/e7c16a1fd92f999807c668a1842fbb59fbb315ee/rust/alloy-op-evm/src/block/mod.rs#L976

mattsse (approved):

PR #364, opened 2026-05-22
feat: handle non-fatal EVM errors

klkvr:
## Motivation

Adds support for non-fatal database errors like the ones returned from BAL database. Those errors are now mapped to `BlockValidationError` variant, allowing to not fail entire block's execution

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/evm/src/block/error.rs
+++ b/crates/evm/src/block/error.rs
@@ -160,13 +160,16 @@ impl BlockExecutionError {
     /// Handles an EVM error occurred when executing a transaction.
     ///
     /// If an error matches [`EvmError::InvalidTransaction`], it will be wrapped into
-    /// [`BlockValidationError::InvalidTx`], otherwise into [`InternalBlockExecutionError::EVM`].
+    /// [`BlockValidationError::InvalidTx`], otherwise if [`EvmError::is_fatal`] return `true`, the
+    /// error would be wrapped into [`InternalBlockExecutionError::EVM`], otherwise it would be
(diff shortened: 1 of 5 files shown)


crates/evm/src/block/error.rs
mattsse, line 166: needs doc updates

crates/evm/src/block/error.rs
```diff
@@ -157,7 +167,11 @@ impl BlockExecutionError {
                 Self::Validation(BlockValidationError::InvalidTx { hash, error: Box::new(err) })
             }
             Err(err) => {
-                Self::Internal(InternalBlockExecutionError::EVM { hash, error: Box::new(err) })
+                if err.is_fatal() {
+                    Self::Internal(InternalBlockExecutionError::EVM { hash, error: Box::new(err) })
+                } else {
+                    Self::Validation(BlockValidationError::EVM { hash, error: Box::new(err) })
```
mattsse, line 176: can we add a oneline doc here about what this is

pepyakin (approved):
sgtm

crates/evm/src/error.rs
```diff
@@ -90,4 +97,12 @@ where
             err => Err(err),
         }
     }
+
+    fn is_fatal(&self) -> bool {
+        match self {
+            Self::Transaction(_) => false,
```
mattsse, line 103: why is this not fatal?

mattsse (approved):

PR #372, opened 2026-06-02
fix: keep moved precompiles cold

Soubhik-10:


mattsse (commented):
can you explain again what this is trying to fix and why the current move impls is insufficient?

PR #380, opened 2026-06-11
fix(evm): enforce EIP-8037 tx.gas <= state_gas_available inclusion check

rakita:
## Description

EIP-8037 specifies the per-dimension block inclusion check as:

    min(TX_MAX_GAS_LIMIT, tx.gas) <= regular_gas_available
    tx.gas <= state_gas_available

`EthBlockExecutor::execute_transaction_without_commit` only enforced the regular-dimension check, so a transaction with `tx.gas > block.gas_limit` could be included whenever the EIP-7825 cap masked it (e.g. `tx.gas = 20_000_001` with a `16_777_216` cap and a 20M block gas limit passes the capped regular check but must be rejected on the state dimension).

This adds the missing state-dimension check, gated on `enable_amsterdam_eip8037`: reject when `tx.gas_limit() > block.gas_limit - block_state_gas_used`, reusing the existing `TransactionGasLimitMoreThanAvailableBlockGas` error. The uncapped gas limit is used deliberately — per the EIP, `TX_MAX_GAS_LIMIT` applies only to the regular dimension.

## Tests

Adds three executor-level unit tests:
- rejection when `tx.gas` exceeds the block gas limit despite the cap satisfying the regular check,
- rejection against remaining state gas after `block_state_gas_used` accrues,
- acceptance exactly at the state-gas boundary.

Verified the rejection tests fail without the fix.

Ref: ethereum/execution-specs#2892 (boundary tests for this rule across tx types).

### DIFF
--- a/crates/evm/src/eth/block.rs
+++ b/crates/evm/src/eth/block.rs
@@ -191,6 +191,21 @@ where
             .into());
         }
 
+        // Amsterdam+ (EIP-8037): the full transaction gas limit must also fit within the
+        // block's available state gas: tx.gas <= state_gas_available. The transaction gas
+        // limit cap applies only to the regular dimension, so the uncapped gas limit is
+        // used here.
+        if self.evm.cfg_env().enable_amsterdam_eip8037 {
+            let state_gas_available = self.evm.block().gas_limit() - self.block_state_gas_used;
+            if tx.tx().gas_limit() > state_gas_available {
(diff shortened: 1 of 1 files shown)


rakita:
Merged already on main https://github.com/alloy-rs/evm/blob/59e1c381b512c5cb81ae0036edf1a30e1bde5c50/crates/evm/src/eth/block.rs#L222-L235

PR #396, opened 2026-07-23
chore: bump revm to 42.0.1 + EIP-8282 system calls

rakita:
Bumps `revm` to `42.0.1` (revm [v115](https://github.com/bluealloy/revm/releases/tag/v115)) and ports the Amsterdam system calls that were previously only on the `glamsterdam-devnet-7` branch, with the devnet-7 predeploy addresses.

### revm bump
No source changes were required for the bump itself.

### EIP-8282 builder execution requests
Two post-block system calls, active from Amsterdam, each called as `SYSTEM_ADDRESS` with empty calldata; their output is pushed as EIP-7685 requests:
- builder deposit requests (type `0x03`) — `0x0000BFF46984E3725691FA540A8C7589300D8282`
- builder exit requests (type `0x04`) — `0x000064D678505AD48F8CCB093BC65613800E8282`

These are the current devnet-7 predeploy addresses; they match revm (`bins/revme/src/cmd/blockchaintest/post_block.rs`) and supersede the earlier `…008282` / `…008282` pair. Two matching `BlockValidationError` variants were added.

### ~EIP-7997 deterministic CREATE2 factory~
Dropped per review: the implemented design is obsolete — EIP-7997 now uses the existing factory at `0x4e59b44847b379578588920cA78FbF26c0B4956C` and no longer performs a fork-boundary state transition. The factory is provided via a normal deployment transaction or genesis configuration instead.

### MSRV bump to 1.94.1
Unrelated to revm: `alloy-consensus`/`alloy-eips`/`alloy-tx-macros` `2.2.0` were published with `rust-version = 1.94.1`. Since the workspace requires `^2.0.0` and no lockfile is committed, CI resolves 2.2.0 and the old 1.91 MSRV jobs fail — `main` is affected too. Bumped `rust-version` and the CI matrix to match. Happy to split this into its own PR if preferred.

Clippy clean; MSRV build verified against a real 1.94.1 toolchain.


crates/evm/src/block/system_calls/eip7997.rs
```diff
@@ -0,0 +1,204 @@
...
+use alloy_hardforks::EthereumHardforks;
+use alloy_primitives::{address, b256, bytes, Address, Bytes, B256};
+use revm::{
+    context::Block,
+    state::{Account, AccountStatus, Bytecode, EvmState},
+    Database,
+};
+
+/// Address of the EIP-7997 deterministic `CREATE2` factory predeploy.
+pub const FACTORY_ADDRESS: Address = address!("0x0000000000000000000000000000000000000012");
```
klkvr: Could we drop this module and the `apply_factory_predeploy` hook? This implements an obsolete EIP-7997 design. Before this PR opened, the EIP had switched to the existing factory at `0x4e59b44847b379578588920cA78FbF26c0B4956C`, different runtime code, and a nonzero nonce. It subsequently removed the fork-boundary lookup/mutation and now says clients MUST NOT check for the factory at the fork boundary. As written, Amsterdam execution writes unrelated code at `0x12`, potentially with nonce zero, causing a consensus state-root divergence. The factory should instead be provided by the normal deployment transaction or genesis configuration.

References: [current specification](https://eips.ethereum.org/EIPS/eip-7997#specification), [factory change](https://github.com/ethereum/EIPs/commit/a0a7f5adb491fc6ad4b008f307899c30f348db22), [irregular state transition removal](https://github.com/ethereum/EIPs/commit/e239a2e8a2a7f4100f5c6605805ae0ca8624b5b9).

rakita: Good catch — removed the module and the `apply_factory_predeploy` hook in 5464284.

rakita:
@klkvr this is ready, factory deployment was removed

klkvr (approved):

PR #399, opened 2026-08-11
fix: enforce block state-gas capacity per check_block_gas_capacity

rakita:
## Summary

Amsterdam's block capacity check has two dimensions per execution-specs `check_block_gas_capacity` (`forks/amsterdam/vm/gas.py`):

1. **Execution dimension** (already implemented): `min(TX_MAX_GAS_LIMIT, tx.gas)` must fit within `block_gas_limit − block_regular_gas_used`.
2. **State dimension** (added here): the **full, uncapped** `tx.gas` must fit within `block_gas_limit − block_state_gas_used`. The `tx_gas_limit_cap` does not bound this dimension — state gas is drawn from the EIP-8037 reservoir that sits *above* the cap.

Previously only the first check ran, so a transaction with a gas limit above the 16.7M cap could be admitted into a block whose state-gas budget was nearly exhausted — accepted here, rejected by the reference spec.

The new check reuses the existing `TransactionGasLimitMoreThanAvailableBlockGas` error variant with `block_available_gas` set to the remaining state-gas budget, keeping the error enum stable.

## Testing

- `cargo nextest run --workspace`: 55/55 pass
- clippy (all targets, all features): clean

Gated on `enable_amsterdam_eip8037`; pre-Amsterdam behavior is unchanged.

### DIFF
--- a/crates/evm/src/eth/block.rs
+++ b/crates/evm/src/eth/block.rs
@@ -197,6 +197,21 @@ where
             .into());
         }
 
+        // Amsterdam+: the state-gas dimension has its own budget of `block_gas_limit`
+        // (execution-specs `check_block_gas_capacity`). Unlike the execution dimension,
+        // the transaction's full gas limit counts against it — state gas is drawn from
+        // the reservoir above `tx_gas_limit_cap`, so the cap does not bound it.
+        if self.evm.cfg_env().enable_amsterdam_eip8037 {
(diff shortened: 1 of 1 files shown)


rakita:
This can be merged right away, it is part of devnet-7/8. Flag is needed for tempo as we want to skip the state gas check, as state gas usage is unbounded in tip-1016

PR #400, opened 2026-08-13
feat(system-calls): reject blocks when EIP-8282 builder contracts have no code

spencer-tb:
A block executed while an EIP-8282 builder predeploy has no code must be invalid, mirroring the EIP-7002/7251 semantics. The system call to an empty contract currently no-ops (empty output, no requests), so such blocks fail much later with misleading errors — on hive glamsterdam-quick (fixtures `tests-glamsterdam-devnet@v8.1.0`), reth rejects the four `test_builder_{deposit,exit}_contract_deployment[deploy_after_fork…]` cases with a BAL hash mismatch where the tests expect `SYSTEM_CONTRACT_EMPTY`; besu/nethermind/erigon/ethrex implement the empty-code check and pass.

Fix: check the predeploy's code before the system call in `transact_builder_{deposit,exit}_requests_contract_call` and return a new `BlockValidationError::SystemContractEmpty { address }` ("system contract {address} has no code") when it is missing. The message shape is deliberately distinct from the `failed to apply … requests contract call` family so downstream consumers (e.g. the EELS exception mapper) can distinguish empty-contract from call-failure rejections.

Verified: `alloy-evm` tests pass; reth (glamsterdam-devnet-8) compiles and its `payload_processor::bal` suite passes against this branch. Companion reth pin: paradigmxyz/reth PR (linked below).

### DIFF
--- a/crates/evm/src/block/error.rs
+++ b/crates/evm/src/block/error.rs
@@ -100,6 +100,14 @@ pub enum BlockValidationError {
         /// The error message.
         message: String,
     },
+    /// A required system contract has no code deployed [EIP-8282]
+    ///
+    /// [EIP-8282]: https://eips.ethereum.org/EIPS/eip-8282
+    #[error("system contract {address} has no code")]
+    SystemContractEmpty {
+        /// The system contract address.
+        address: alloy_primitives::Address,
+    },
     /// Error when decoding deposit requests from receipts [EIP-6110]
     ///
     /// [EIP-6110]: https://eips.ethereum.org/EIPS/eip-6110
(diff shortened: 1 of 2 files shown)


BarneyChambers:
Hey @spencer-tb hope you don't mind that I reviewed your code here and have a small finding:

`eip7002.rs` and `eip7251.rs` on this branch still go straight into `transact_system_call`. `ensure_contract_deployed` only runs for the two 8282 addresses, so an empty 7002 or 7251 still comes back as a successful stop.

do we need to call `ensure_contract_deployed` from `eip7002.rs` and `eip7251.rs` too, before transact_system_call? That function reads the address and rejects the block when the account has no code. This pull request only calls it for the two 8282 addresses.

PR #403, opened 2026-08-26
chore: bump revm to 43.0.0

rakita:
Bumps revm from 42.0.1 to the latest published release, 43.0.0.

The devnet-8 changes are included in the published release, so this also removes the obsolete git patch overrides.

Checks:
- cargo fmt --all -- --check
- cargo check --workspace --all-targets --all-features
- cargo clippy --workspace --all-targets --all-features -- -D warnings
- cargo test --workspace --all-features


crates/evm/src/eth/block.rs
```diff
@@ -184,10 +209,7 @@ where
 
         // Use regular part of transaction gas limit to check if it fits inside available block
         // space.
-        let mut max_tx_gas_usage = tx.tx().gas_limit();
-        if let Some(tx_gas_limit_cap) = self.evm.cfg_env().tx_gas_limit_cap {
-            max_tx_gas_usage = min(max_tx_gas_usage, tx_gas_limit_cap);
-        }
+        let max_tx_gas_usage = min(tx.tx().gas_limit(), self.evm.cfg_env().tx_gas_limit_cap());
```
mattsse, line 212: this looks like a driveby

crates/evm/src/eth/block.rs
```diff
@@ -371,7 +408,10 @@ impl<R, Spec, EvmF> BlockExecutorFactory for EthBlockExecutorFactory<R, Spec, Ev
 where
     R: ReceiptBuilder<Transaction: Transaction + Encodable2718, Receipt: TxReceipt<Log = Log>>,
     Spec: EthExecutorSpec,
-    EvmF: EvmFactory<Tx: FromRecoveredTx<R::Transaction> + FromTxWithEncoded<R::Transaction>>,
+    EvmF: EvmFactory<
+        Spec: Into<SpecId> + Clone,
+        Tx: FromRecoveredTx<R::Transaction> + FromTxWithEncoded<R::Transaction>,
+    >,
```
mattsse, line 414: why do we need this now

crates/evm/src/eth/block.rs
```diff
@@ -144,7 +165,11 @@ where
 
 impl<E, Spec, R> BlockExecutor for EthBlockExecutor<'_, E, Spec, R>
 where
-    E: Evm<DB: StateDB, Tx: FromRecoveredTx<R::Transaction> + FromTxWithEncoded<R::Transaction>>,
+    E: Evm<
+        DB: StateDB,
+        Spec: Into<SpecId> + Clone,
+        Tx: FromRecoveredTx<R::Transaction> + FromTxWithEncoded<R::Transaction>,
+    >,
```
mattsse, line 172: same q

crates/evm/src/eth/block.rs
klkvr: yeah it was written like that to avoid `Spec` bound which doesn't hurt but still seems ok to keep as is?

klkvr (approved):

crates/evm/src/eth/block.rs
rakita: Yeah exactly what @klkvr said, it is bcs of spec. ~~We now need to know fork, and if it is amsterdam or not.~~ 

Not exactly for fork but for default value.

PR #408, opened 2026-09-18
chore(ci): route package installs through Aegis

decofe:
## Motivation

Extend the Aegis rollout to all applicable workflows in this repository.

## Solution

Initialize the same SHA-pinned secure runner as [alloy#4203](https://github.com/alloy-rs/alloy/pull/4203) before checkout and downloads, with job-scoped `id-token: write`. Review all 1 workflow files and 12 job definitions.

Run cargo-deny directly on the protected host instead of the reusable/container action, preserving all-features checks. Pin tool versions in the vendored installer and keep checksum verification enabled.

## Validation and limits

YAML parsing, first-step/OIDC coverage assertions, and `git diff --check` passed. Actionlint reports no new diagnostics compared with the default branch. Full platform execution and live policy enforcement remain subject to GitHub CI.

The action configures Cargo, npm/pnpm/Yarn/Bun, Go, pip/uv/Poetry, and RubyGems/Bundler. Fork PRs without OIDC skip package enforcement with a warning. Cached artifacts, prebuilt tools, Docker images, OS packages, and direct binary downloads are not per-package policy checks. Existing dependency/lockfile policy is preserved.

Prepared with AI assistance.

Prompted by: @grandizzy


.github/workflows/ci.yml
```diff
@@ -270,14 +334,20 @@ jobs:
...
         with:
           persist-credentials: false
-      - uses: taiki-e/install-action@1ed6d7be6168f6c9046541087ff549b6bc581fdf # v2
+      - name: Install zepter
+        uses: tempoxyz/gh-actions/vendor/taiki-e/install-action@25cce154e7fb10f99361a166468a6c56b9c31aa3
         with:
-          tool: zepter
+          tool: zepter@1.88.1
+          checksum: true
+          fallback: cargo-binstall
```
mablr, line 350: `zepter` has no manifest in this pinned installer, so this always takes the cargo-binstall fallback https://github.com/tempoxyz/gh-actions/blob/25cce154e7fb10f99361a166468a6c56b9c31aa3/vendor/taiki-e/install-action/main.sh#L1093-L1107. `checksum: true` verifies the cargo-binstall bootstrap, but the fallback does not require a checksum/signature for the final tool binary. We should probably use an exact-version locked source install through Aegis, or add verified artifact support in `gh-actions` and disable fallback.

grandizzy: Addressed in 8f159d4 by replacing the cargo-binstall fallback with `cargo install zepter --version 1.88.1 --locked` after Secure Runner, so the exact-version locked source install is routed through Aegis. [CI passed](https://github.com/alloy-rs/evm/actions/runs/35573604671). The install step took 52s versus 2s in the previous run (+50s); feature-propagation took 2m00s versus 48s, but still completed 1m51s before the final required test job, so it did not affect the critical path in this run.

mablr (commented):
LGTM

grandizzy (approved):

PR #411, opened 2026-09-22
fix(evm): preserve fatal system call error sources

banteg:
Fatal database errors during system calls are currently converted to validation-error strings, losing their cause type and treating unavailable state as an invalid block. Preserve them as internal execution errors with the original source chain and diagnostic context; nonfatal errors and revert/halt handling keep their existing behavior.

Six real-EVM regressions fail before the fix and pass afterward, covering blockhash, beacon-root, withdrawal, consolidation, builder-deposit and builder-exit calls. Workspace tests pass with default and no default features (61 unit tests and one doctest each; 12 existing ignored doctests), along with nightly formatting and all-target/all-feature Clippy.

Found while following up [trace-interop's pruned-history results](https://github.com/banteg/trace-interop/blob/main/evidence/2026-09-22/reth-main-534c60db/README.md), part of the [trace API standardization work](https://github.com/ethereum/execution-apis/issues/890). Tracing maintainers are welcome in the [working group](https://t.me/+2jfwc-YvwkliMzIy).


banteg:
@klkvr could you approve the pending fork workflows and review this when convenient? The six system-call error-source regression tests passed locally, but upstream CI is still approval-gated. Reth's corresponding error mapping is tracked in paradigmxyz/reth#27378; the complete fix also needs dependency uptake there.

